PR #9, opened 2020-11-08
Add e2e tests to github action testing

addshore:


AlaaSarhan:
Locally it runs successfully now, as long as you have firefox installed

I removed chrome as making that run locally requires also installing chromedriver and keeping it up-to-date with chrome version. It would probably be nice to have a docker container do that for us, if tests against chrome will be really needed locally.

PR #21, opened 2020-11-08
Add basic UI for equivEntites

### LINKED ISSUE #16: Ability to configure wgWbManifestWikidataEntityMapping
When the wikis end up on 1.35 (should be very soon) WikibaseManifest will then be deployed.
https://github.com/wikimedia/mediawiki-extensions-WikibaseManifest

This extension requires users to be able to configure some equivalent entities.
This needs to be exposed in the wbstack UI somehow.
API to come, but will look similar to the other APIs that exist for mediawiki / wikibase settings.

tarrow:
fixes #16

TODO:
- some validation
- make sure pushed data format is compatible with the api


addshore:
Thanks for the draft!
It looks like this will probably now be tackled as part of https://github.com/wbstack/ui/pull/28 so going to close this for now

PR #32, opened 2021-03-24
Add settings sections for mapping entities

tzhelyazkova:
This adds a section for mapping properties and a section for mapping items.
It's using vuetify's `data-table` with crud actions.

Current problem: can't make the actions show up in the UI. 🤔

### DIFF
--- a/src/components/Pages/ManageWiki/Features/ItemsTable.vue
+++ b/src/components/Pages/ManageWiki/Features/ItemsTable.vue
@@ -1,0 +1,222 @@
+<template>
+<div>
+  <v-data-table
+    :headers="headers"
+    :items="wikibaseItems"
+    sort-by="wikidata"
+    class="elevation-1"
+  >
+    <template v-slot:top>
+      <v-toolbar
+        flat
+      >
+        <v-toolbar-title>Items</v-toolbar-title>
+        <v-divider
+          class="mx-4"
+          inset
+          vertical
+        ></v-divider>
+        <v-spacer></v-spacer>
+        <v-dialog v-model="dialogDelete" max-width="500px">
+          <v-card>
+            <v-card-title class="headline">Are you sure you want to delete this mapping?</v-card-title>
+            <v-card-actions>
+              <v-spacer></v-spacer>
+              <v-btn color="blue darken-1" text @click="closeDelete">Cancel</v-btn>
+              <v-btn color="blue darken-1" text @click="deleteItemConfirm">OK</v-btn>
+              <v-spacer></v-spacer>
+            </v-card-actions>
+          </v-card>
+        </v-dialog>
+      </v-toolbar>
+    </template>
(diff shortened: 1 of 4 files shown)


src/components/Pages/ManageWiki/Features/PropertiesTable.vue
```diff
@@ -0,0 +1,224 @@
...
+    created () {
+      this.initialize()
+    },
+
+    methods: {
+      initialize () {
+        this.properties = [
+          {
+            local: 'No Mapping',
+            wikidata: 'P31 (Instance of)',
```
jakobw, line 169: This would somehow need to be serialized as just "P31" when we send it to the API right? So we need to either make it possible to display something other than the id value (i.e. keep both label and value here) or change this to only be 'P31' :thinking:

src/components/Pages/ManageWiki/Features/ItemsTable.vue
```diff
@@ -0,0 +1,216 @@
...
+		>
+		  <template v-slot:activator="{ on, attrs }">
+			<v-btn
+			  color="primary"
+			  dark
+			  class="mb-2"
+			  v-bind="attrs"
+			  v-on="on"
+			>
+			  Add New Mapping
```
jakobw, line 31: Not sure if this really matters but in the mockups the buttons are below the tables :man_shrugging:

jakobw (commented):
Looks fancy! :sparkles: 

ItemsTable and PropertiesTable are pretty similar and look like they should share some code, but it's probably ok to look into that later. It might turn out that things like different validation becomes harder if they're not duplicated.

My IDE is really unhappy about using tabs in some of these components but oh well.

src/components/Pages/ManageWiki/Features/PropertiesTable.vue
tzhelyazkova: Right, I had the same thought.
We should either ask UX if it's ok to just omit the label part and only say P31 or we should somehow cut it out before passing it to the api.

src/components/Pages/ManageWiki/Features/ItemsTable.vue
tzhelyazkova: Moved it

src/components/Pages/ManageWiki/Features/PropertiesTable.vue
tzhelyazkova: an executive decision was taken to remove that bit and leave only the id 😄

addshore:
It looks like I'm able to add properties into the items table, and items into the properties table.
So I guess we should add validation to the TODOs :)

addshore:
Spotted another things.
The sorting is unfortunately alphabetical :( would be cool if it was numerical based on everything except the first character
image

PR #50, opened 2021-03-24
save manifest-mapping

toban:


toban:
"we probably want to keep the entity mapping in the store's state. that way we don't have to worry about peeling it out of the component refs (usually a bad idea), and easily sync it from the store to the component on load"

from what jacob said

src/components/Pages/ManageWiki/Tabs/ManageFeatures.vue
```diff
@@ -2,7 +2,7 @@
   <v-content>
     <FederatedProperties :wikiId="this.wikiId" />
     <v-divider class="space"></v-divider>
-    <EntityMapping/>
+    <EntityMapping :wikiId="this.wikiId" />
```
tzhelyazkova, line 5: just `wikiId`

src/components/Pages/ManageWiki/Features/EntityMapping.vue
```diff
@@ -24,20 +24,43 @@
...
   name: 'EntityMapping',
   components: { PropertiesTable, ItemsTable },
-  props: [],
+  props: ['wikiId'],
   methods: {
+    toMapping (entities) {
+      const mappings = []
+      for (var i = 0; i < entities.length; i++) {
+        const mapping = entities[i]
+        mappings[mapping.local] = mapping.wikidata
```
tzhelyazkova, line 45: I think it should be the opposite `mappings[mapping.wikidata] = mapping.local` because the tools that read the setting should be parsing it by wikidata's IDs.

src/components/Pages/ManageWiki/Features/EntityMapping.vue
```diff
@@ -24,20 +24,43 @@
...
+      }
+      return mappings
```
tzhelyazkova, line 47: Manifest actually splits the array in properties and items, not everything all together.
example from my `LocalSettings.php`:
```
$wgWbManifestWikidataEntityMapping = [
    'properties' => [
        "P31" => 'P1'
    ],
    'items' => [
        'Q42' => 'Q1'
    ]
];
```
We should reflect it here, either with some check of what is passed or maybe split it in two functions, whatever sounds easier.

tzhelyazkova (commented):
I think this is pretty much the way to go it just needs a tweak to the final mappings array (see comments below).

P.S. For some reason I couldn't add suggestions but only comments :/

src/components/Pages/ManageWiki/Features/EntityMapping.vue
toban: done!

src/components/Pages/ManageWiki/Features/EntityMapping.vue
toban: done! https://github.com/wbstack/api/pull/35

src/components/Pages/ManageWiki/Features/ItemsTable.vue
```diff
@@ -191,6 +197,8 @@ export default {
 
     deleteItemConfirm () {
       this.wikibaseItems.splice(this.editedIndex, 1)
+      // TODO: this way of deleting an item pair is silly. how to do computed setters for arrays?!
```
jakobw, line 200: I think the answer to this smart comment that I wrote is: don't do computed setters for arrays. (Kind of makes sense: if it's mutated by calling a method on the same object, it's not a setter. duh!)

So the right way :tm: to do this is probably to just have a `wikibaseItems() { /* getter code */ }` computed property (no `get`/`set` pair) and instead of using a setter we write to the store through separate `addItem`, `deleteItem`, `editItem` methods for which we can either create separate mutations or have them call the `setItemMapping` as we do now in the setter.

addshore:
When testing this against the real API the list of wikis on `/dashboard` does not appear to populate when logging in any more.
Reloading the app once logged in though seems to work.
This was tested on https://4c30ae7678-prod--wbstack-ui.netlify.app/ which points to the prod api

src/components/Pages/ManageWiki/Features/PropertiesTable.vue
```diff
@@ -168,24 +182,7 @@ export default {
     }
   },
 
-  created () {
-    this.initialize()
-  },
-
   methods: {
-    initialize () {
```
tzhelyazkova, line 176: hmm I think this should stay, because the mocks show suggesting properties to the user in this way.
So if there are no equiv entities settings already, we show these suggestions.

tzhelyazkova: Oh, nvm https://github.com/wbstack/ui/pull/67 does that.

jakobw: I uploaded #67 where they're initialized in the store instead

addshore:
> > When testing this against the real API the list of wikis on /dashboard does not appear to populate when logging in any more.
> > Reloading the app once logged in though seems to work.
> > This was tested on https://4c30ae7678-prod--wbstack-ui.netlify.app/ which points to the prod api
> 
> @addshore I think this and the login bug you found on #65 were both fixed with [581cb71](https://github.com/wbstack/ui/commit/581cb71d46b454190b052855b5baaec46ae61d15). At least it looks like the wiki list populates fine for me on the prod version.

looks good

PR #60, opened 2021-03-25
Validate Entity ids

jakobw:


src/components/Pages/ManageWiki/Features/ItemsTable.vue
```diff
@@ -142,7 +148,8 @@ export default {
     defaultItem: {
       local: '',
       wikidata: ''
-    }
+    },
+    isValidItemId: (val) => /^Q\d+$/.test(val) || `"${val}" is not a valid Item id`
```
toban, line 152: This allows for Q0 for example, which is a valid entity id however would probably never exist? I don't think we care do we?

toban (approved):
LGTM and works as expected

src/components/Pages/ManageWiki/Features/ItemsTable.vue
addshore: Sounds like something not worth caring about!

PR #62, opened 2021-03-25
Add a tooltip about saving taking a few seconds

tzhelyazkova:


addshore:
To me `Please note saving your changes takes up to 10 seconds` sounds like the user might be waiting on this page for 10 seconds after clicking the button.
Perhaps something like `it may take up to 10 seconds for changes to be reflected on your wiki` might work

tzhelyazkova:
> it may take up to 10 seconds for changes to be reflected on your wiki

done 👍

PR #64, opened 2021-03-25
Refactor: Unify Property and Item tables into EntityTable

toban:


tzhelyazkova (commented):
If this is the preferred way to go then might as well delete `ItemsTable` and `PropertiesTable` and directly use `EntityTable` in `EntityMapping.vue`

src/components/Pages/ManageWiki/Features/EntityTable.vue
```diff
@@ -0,0 +1,246 @@
...
+        </v-card-actions>
+        </v-card>
+    </v-dialog>
+</div>
+</template>
+
+<script>
+export default {
+  name: 'EntityTable',
+  props: ['mappingName', 'entityValidation', 'entityName', 'storeAction'],
```
tzhelyazkova, line 129: I think passing a store action name to a component is a code smell.
The component should either be aware of the store or not. And here we have it work with the store but we don't tell it exactly in what context so it's a bit strange. 
Ultimately, this component would then not use the store and just communicate to a parent via events.

tzhelyazkova (commented):
I understand the feeling of wanting to refactoring these 2 components because they look pretty similar in the UI but I don't see a big benefit currently or at least until we know we can have a 3rd type of entity mapped.

src/components/Pages/ManageWiki/Features/EntityMapping.vue
```diff
@@ -1,45 +1,70 @@
...
-import ItemsTable from './ItemsTable.vue'
+import EntityTable from '../Features/EntityTable.vue'
 
 export default {
   name: 'EntityMapping',
-  components: { PropertiesTable, ItemsTable },
-  props: ['wikiId']
+  components: { EntityTable },
+  props: ['wikiId'],
+  computed: {
```
toban, line 46: Brilliant!

toban (commented):
Looks great and seems to work! (Can't approve or merge)

jakobw (approved):
Awesome! :rocket:

PR #65, opened 2021-03-25
Prevent errors if `wikibaseManifestEquivEntities` doesn't exist

jakobw:
This prevents an error in the `set_current_wiki_settings` mutation in case the `wikibaseManifestEquivEntities` setting isn't set.

It also adds `public_settings: []` to the mock wiki data. You'll need to clear local storage for it to take effect.

### DIFF
--- a/src/backend/mocks/default_handlers.js
+++ b/src/backend/mocks/default_handlers.js
@@ -35,7 +35,8 @@ const makeNewWiki = ({ domain, sitename }) => {
       id: 101,
       wiki_id: lastWikiId,
       version: 'mw1.33-wbs1'
-    }
+    },
+    public_settings: []
   }
 
   myWikis.push(newWiki)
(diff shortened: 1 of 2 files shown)


addshore:
https://7c7ad5c518-prod--wbstack-ui.netlify.app seems to direct me straight to `/login` when loading the app?
Doesn't appear to happen in the mocked site.
It looks like this must be a bug introduced in this patch though

addshore:
I think this is because the `xhr.js:177 POST https://www.wbstack.com/api/wiki/mine 401` request happens before the user is logged in?
This then forces them to the login page?
But this request shouldn't happen on the homepage

jakobw:
> https://7c7ad5c518-prod--wbstack-ui.netlify.app seems to direct me straight to `/login` when loading the app?
> Doesn't appear to happen in the mocked site.
> It looks like this must be a bug introduced in this patch though

This happened in #50 which this PR is based on with https://github.com/wbstack/ui/pull/50/files#diff-7a7a37b12ee1265d7e27577ec286120d2cbc0940908635e264a2be44ccb9a8a0R45 - loading wikis in `App.vue` without knowing if the user is logged in or not sounds silly in hindsight :D

PR #67, opened 2021-03-25
Re-add default Properties for equiv entities

jakobw:


tzhelyazkova (approved):
👍

addshore:
`value: ["Value must be a JSON string mapping Wikidata Item or Property Ids to local Item or Property Ids"]`
as we are sending `No Value` (we send them both when only 1 is provided)

PR #69, opened 2021-03-25
Wait for settings to be initialized

toban:
adds spinning thing on top of page instead of breaking

### DIFF
--- a/src/components/Pages/ManageWiki/TabSettings.vue
+++ b/src/components/Pages/ManageWiki/TabSettings.vue
@@ -1,5 +1,6 @@
 <template>
     <div>
+      <div v-if="isReady">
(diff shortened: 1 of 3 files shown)


addshore:
CI should be fixed after https://github.com/wbstack/ui/pull/71

addshore (changes requested):
This seems to work for the very initial load of the manage page for a single wiki.
however if you navigate back to the dashboard and select a new wiki you will not see the spinner again

toban:
> This seems to work for the very initial load of the manage page for a single wiki.
> however if you navigate back to the dashboard and select a new wiki you will not see the spinner again

clearing it before making the request now.

addshore (approved):

PR #76, opened 2021-03-26
toggle fedprops

toban:


addshore:
Tested and it functionally seems to do what is expected, though it also still says `[WIP]` currently
I wonder too if there would be any way to make the difference between the 2 options (disable & enable) more obvious.
That is probably a question more for UX though...

toban:
> Tested and it functionally seems to do what is expected, though it also still says `[WIP]` currently
> I wonder too if there would be any way to make the difference between the 2 options (disable & enable) more obvious.
> That is probably a question more for UX though...

There was an issue when no settings was present but should be fixed now. Agree that the difference isn't quite obvious. However turning it back off is equally dangerous as turning it back on i guess ;)

src/components/Pages/ManageWiki/Features/FederatedProperties.vue
```diff
@@ -58,18 +63,22 @@ export default {
...
   methods: {
     enableFeature () {
       const wiki = this.wikiId
       const setting = 'wikibaseFedPropsEnable'
-      const value = 1
+      const value = !this.isEnabled
 
       this.$store.dispatch('updateSetting', { wiki, setting, value })
-        .then(() => this.hideModal())
+        .then(() => this.$store.dispatch('setFederatedPropertiesEnabled', value))
```
jakobw, line 78: The store has this information so we could as well just dispatch a "toggleFederatedPropertiesEnabled" action without any value. This is fine too, though.

jakobw (approved):
Looks great!

(Also I'm slow and this is merged already. Yay! :D)

PR #78, opened 2021-03-26
Sync skin setting with store

toban:
- use it in the settings ui


addshore (approved):
Looks good and seems to work, I wonder if we should have the same tooltip as for other settings saying it may take 10 seconds to get to the wiki?

PR #79, opened 2021-03-26
Sync wikibase string lengths with store

toban:


src/store/wikis.js
```diff
@@ -58,10 +58,22 @@ const mutations = {
     const federatedPropertiesSetting = details.public_settings.find(setting => setting.name === 'wikibaseFedPropsEnable')
     const wikibaseFedPropsEnable = federatedPropertiesSetting ? parseInt(federatedPropertiesSetting.value) === 1 : false
 
+    const wwWikibaseStringLengthStringSetting = details.public_settings.find(setting => setting.name === 'wwWikibaseStringLengthString')
```
toban, line 61: this looks like puke

src/components/Pages/ManageWiki/Cards/Wikibase.vue
```diff
@@ -68,18 +85,17 @@ export default {
...
 
       Promise.all(promises)
         .then(() => {
           alert('Update success!')
-          this.$router.go()
         })
         .catch(err => {
           console.log(err.response)
           alert('Something went wrong.')
-          this.$router.go()
```
addshore, line 82: Glad to be getting rid of these fun terrible reloads :D

src/store/wikis.js
addshore: I think at some point we would want to make the API return the default values.
Then most of this puke can go and just use whatever the API would send.
This puts the control of the defaults etc in the API as well which is probably where it belongs.

addshore (approved):
Seems to be working!
(but could also do with a tooltip about update lag)

jakobw (approved):
LGTM :sparkles:

PR #80, opened 2021-03-26
Sync wgLogo setting with the store

jakobw:
Needs https://github.com/wbstack/api/pull/38 to work


src/components/Pages/ManageWiki/Cards/Logo.vue
```diff
@@ -11,6 +11,7 @@
         :persistent-hint="true"
         @change="onLogoFileChanged"
       ></v-file-input>
+      <v-img v-if="logo" :src="logo" max-width="135" />
```
toban, line 14: Hmm when I'm testing with the suggested hack in https://github.com/wbstack/api/issues/37 I'm still not seeing this however the logo is there so Im guessing the upload just didn't work. Seems to read / write it successfully from db though.

src/store/wikis.js
```diff
@@ -55,9 +55,13 @@ const mutations = {
...
     const wikibaseFedPropsEnable = federatedPropertiesSetting ? parseInt(federatedPropertiesSetting.value) === 1 : false
 
+    const logoSetting = details.public_settings.find(setting => setting.name === 'wgLogo')
+    const logoUrl = logoSetting ? logoSetting.value : null
+
     state.currentWikiSettings = {
       entityMapping,
-      wikibaseFedPropsEnable
+      wikibaseFedPropsEnable,
+      logoUrl
```
toban, line 64: I kept naming these with the same name as they are stored in the db to avoid any [confusion](https://github.com/wbstack/api/blob/main/app/Http/Controllers/WikiSettingController.php#L23). entityMapping isn't so maybe we don't care?

addshore:
image
 :D

addshore:
> Needs [wbstack/api#38](https://github.com/wbstack/api/pull/38) to work

This is merged and deployed

addshore:
Followup! https://github.com/wbstack/ui/pull/84

src/store/wikis.js
jakobw: Keeping these names aligned does seem like a good idea. I'll upload a follow-up, thanks!!

PR #98, opened 2021-03-26
add screenshot after test

toban:
- adds wdio-mediawiki for screenshot function


addshore:
Worked in node 12, but not 10 or 14? :D
Looks like we should add `/home/runner/work/ui/ui/tests/e2e/logs` as github action artifacts!

PR #260, opened 2021-10-19
Read config from config.js instead of env vars

jakobw:
This moves config values away from environment vars and into a dedicated `config.js` file in preparation for a solution similar to what's suggested in https://stackoverflow.com/questions/53010064/pass-environment-variable-into-a-vue-app-at-runtime

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,8 +1,6 @@
 version: '3.3'
 services:
   ui:
-    env_file:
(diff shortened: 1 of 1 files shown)


tarrow:
I believe this might need some alterations to `docker-compose.yml`. Right now it still references the missing `.env` file and doesn't seem to mount in any runtime config

PR #390, opened 2022-04-01
Remove old PM and text from dashboard

toban:


toban:
image

from the netlify link looks good

PR #435, opened 2022-05-11
Add naming requirements in the hover icon near Site Domain

dati18:
Screenshot from 2022-05-11 14-55-47


src/components/Cards/CreateWiki.vue
```diff
@@ -35,6 +35,7 @@
             <span>A domain name is the site address people type into their browser to visit your site</span><br/>
             <span>If you already own a domain you can use it by selecting "Custom Domain"</span><br/>
             <span>If not you can use a "Free Subdomain"</span><br/>
+            <span>The subdomain must be at least four-character long, and can consist of small Latin letters, digits and hyphens</span><br/>
```
toban, line 38:
```suggestion
            <span>The subdomain must be at least five-character long, and can consist of small Latin letters, digits and hyphens</span><br/>
```

toban (commented):
Looks like what the designer asked for! However i think i found a problem with the statements.

PR #447, opened 2022-05-27
Catch and display error when too many wikis

toban:


src/components/Cards/CreateWiki.vue
```diff
@@ -228,6 +228,10 @@ export default {
         this.hasError = true
         this.error.terms = errors.terms[0]
       }
+      if (errors.tooManyWikis) {
+        this.hasError = true
+        this.displayGenericError(errors.message)
```
tarrow, line 233: This obviously works fine but it "feels" like we would maybe want to TODO: display `errors.message` in almost all cases

PR #497, opened 2022-07-26
Password reset flow

tarrow:


src/components/Cards/ForgottenPassword.vue
```diff
@@ -14,14 +17,21 @@
...
       </v-form>
+      <v-card-actions>
+        <v-spacer></v-spacer>
+        <v-btn @click="reset" color="primary" :disabled="inFlight">{{buttonText}}</v-btn>
+      </v-card-actions>
+      <v-alert class="mt-8 mr-2 ml-2" outlined type="error" border="left" v-if="error">
+        {{error}}
+      </v-alert>
+      <v-alert class="mt-8 mr-2 ml-2" outlined type="success" border="left" v-if="success">
+        An email has been sent to <b>{{email}}</b>
```
toban, line 30: This message is shown if I enter anything or nothing. Shouldn't there be at least a check that what is entered wasn't empty and possibly an email?

src/components/Cards/ForgottenPassword.vue
```diff
@@ -60,7 +71,7 @@ export default {
 
       this.$store
         .dispatch('forgottenPassword', { email })
-        .then(() => this.$router.push('/'))
+        .then(() => { this.success = true })
```
toban, line 74: I think this actually makes it a little bit less user friendly as every request now is considered a success, even though the input data isn't accepted by the api.

toban: image

src/components/Cards/ForgottenPassword.vue
dati18: we also talked about it as I remember, but the ticket doesn't mention it and also there's no design to do so

dati18: This could be a follow-up ticket I guess

src/components/Cards/ForgottenPassword.vue
dati18: If you write `() => this.succes = true`, it will cause some problem with linting but resolve the problem, so I have to suppress linting in that line.

tarrow: I stared at this a bit more and fundamentally this logic just just isn't really sound at all. I'm pretty sure we need to be replying on the state *way more* for this all to work exactly as intended

src/components/Cards/ForgottenPassword.vue
tarrow: resolved to check that it is not empty with the rules (and validation on submit) on the `v-text-field`

addshore:
Bye eye the code change looks pretty good

Testing out the mocked UI on netlify...

The user can't appear to alter their email address if they get an error? Instead the button and text box are disabled, while the error shows.



If I navigate away from the failed password reset and back to reset password, the error then still displays, but the content (email) is gone, and the buttons is re enabled.

Interestingly this error then can dynamically change as I retype the email
Not in the screenshot below this error is for a previous request for a different email address.



The happy path status in comparison correctly resets between views of the page



Testing against the real API things don't go as smoothly

addshore:
I think this looks pretty good as is now

addshore (approved):

dati18 (approved):
it looks good to me

PR #536, opened 2022-09-11
Tag docker images with short git SHA

tarrow:


addshore:
Looking at this PR just made me realize that all images that are built on the `main` branch are already pushed to the container registry

image
https://github.com/wbstack/ui/pkgs/container/ui/versions

An alternative tagging option thus would be instead of again tagging all images that are built (perhaps just more work if they all end up in the registry), instead just tag on deploy?
This could be more valuable and give more visibility, but I'm not actually sure how well this integrates into CD applications such as argo.

For example, at any given time images could be tagged as `wikibase.cloud` and `wikibase.dev` for example. But also each deployment at a given time could also tag, `wikibase.cloud-20220911010101` for example.
Just a thought, certin

.github/workflows/docker.build.yml
```diff
@@ -21,6 +21,8 @@ jobs:
         uses: crazy-max/ghaction-docker-meta@v4.0.1
         with:
           images: ghcr.io/${{ github.repository }}
+          tags: |
+            type=sha
```
addshore, line 25: Do we want to tag all things that trigger this job?
Note this means unmerged un reviewed pull requests.

I imagine it might be cleaner to keep the tagged images in the image repository as merged into main code?

tarrow: I might be misunderstanding but this doesn't tag and push images built from random PRs: sure they get tagged but they are never pushed to ghcr (see: https://github.com/wbstack/ui/pull/536/files#diff-c59f4f9a90c4b34cef2c040c881f906daad13fde73566f0e34954f46b0f6efa3R57)

tarrow:
I'm not sure I fully understand but tagging images based on where they have been used rather than "what they are" doesn't sound like it would be doing things in the right order.

Moving towards a CD/CI world should mean us moving towards every commit having an image built, and that image being deployed in all environments. This would enable us to "stop thinking" about picking special images and generally "continuously"  deploy each change we merge

addshore:
Yeah I agree, and I dont think the tagging on deployment would work well with something like argo.

> Moving towards a CD/CI world should mean us moving towards every commit having an image built, and that image being deployed in all environments. This would enable us to "stop thinking" about picking special images and generally "continuously" deploy each change we merge

So thinking about the CI/CD pipeline for this wouldn't it actually be that something like argo is looking for new versions of charts to deploy, not actually images?
ie
 - new commit in the code repo merged
 - image built, tagged and pushed to registry
 - chart updated to use new image, and pushed to registry
 - argo sees and deploys new chart?

There is an alternative with some known issues of course, which would be just deploy the code in an old chart, and fix the chart if things break.

Also looking at argo, I imagine we will use `update-strategy: latest` which I believe means we will also want to use `allow-tags: regexp:<expression>` to avoid any other tags that we may want to push at any point triggering a deploy.
So perhaps we should prefix the tags with `main-` right away?

But also regarding what the next step is, it could be worth us spending a few minuites to try and sketch it up, is it argo, is it something else.
Argo and image updater for example wouldnt get to your desired state of actually deploying every commit merged into main in order, as it just polls for the latest thing after 2 mins or so, so commits can actually be missed.

tarrow:
I think some big up front plan might be nice but I'm not sure this small PR is the place to discuss it.

If we were to use argo (for example) there is no particular reason why we would be watching for new charts and updating those (although of course we could) rather than using a generally fixed chart and altering the value of the image tag passed into it

tarrow:
given that this is quite easy to undo (we can simply remove tags that we're up happy with) maybe we would be happy to try this; particularly given that https://github.com/wmde/wbaas-deploy/blob/main/doc/adr/0005-ui-by-commit-hash.md is now approved and merged

deer-wmde:
> given that this is quite easy to undo (we can simply remove tags that we're up happy with) maybe we would be happy to try this

agreed - lets see if this works! also I think this couldnt really break anything, so lets go

deer-wmde (approved):

addshore:
I'd be very pro prefixing the tags with "main-" out of the gate to make the tags easier to understand when not looking at the job itself that creates the tags.

PR #545, opened 2022-09-30
Frontend for contact page

rosalieper:


tarrow:
Generally looks good to me from a code perspective.

I added a thought to the ticket about which key the recaptcha token is sent in. Right now this is not in the key specified in the ticket but it might be that we want to change the specification not the code.

One thing that is notable is that this doesn't implement some error state. We should probably add that. (see https://phabricator.wikimedia.org/T318400#8428059).

tarrow:
Generally the functionality looks like the mock. The only thing I notice is missing (other than the above error handling) is length limits on the other fields and "required" validation. I think we "agreed" 300 chars would be sensible for these other fields. Probably best not to show a counter on these though :)

dati18:
@tarrow Charlie asked for a counter as I remember?
here: https://phabricator.wikimedia.org/T314570#:~:text=Addition%20regarding%20the,can%20be%20specified

tarrow:
so I think the message field should have both the validation and the counter.

I think the other fields should have the validation (for both being "required" if the are and for the max length) but no counter

src/components/Pages/Contact.vue
```diff
@@ -1,23 +1,153 @@
...
+    ],
+    name: '',
+    contactDetails: '',
+    subject: '',
+    message: '',
+    successMessage: false,
+    errorMessage: false
+  }),
+  methods: {
+    closeAlert () {
```
dati18, line 94: This method will close the snackbar in both case (success and error). I don't know if it'll be better if I have 2 separate methods for each case? @tarrow

tarrow: I think there isn't a massive motivation to go in either direction. Right now I think this is fine :)

tarrow (approved):

PR #585, opened 2022-12-01
Update create account page

rosalieper:


src/components/Pages/CreateAccount.vue
```diff
@@ -1,12 +1,12 @@
...
   <v-container class="fill-height" fluid >
     <v-row align="center" justify="center">
-      <CreateAccountCard title="Lets get started" buttonText="Create Account"/>
+      <CreateAccountCard title="Sign up" buttonText="Create Account"/>
     </v-row>
       <v-row align="center" justify="center">
       <v-col class="needs-padding">
         <p>Already have an account?</p>
-        <router-link to="/login">Login now.</router-link>
+        <router-link to="/login">Log in</router-link>
```
tarrow, line 9: Not sure this should have lost the `.`

rosalieper: I changed it to a btn as per the AC.

src/components/Cards/CreateAccount.vue
```diff
@@ -4,10 +4,16 @@
       <v-toolbar-title>{{title}}</v-toolbar-title>
     </v-toolbar>
     <v-card-text>
+      <p>
+        Wikibase.cloud is in a closed beta. If you'd like to receive an invite code,
+        <a href="https://lime.wikimedia.de/index.php/717538">
```
dati18, line 9: I think we should add `target="_blank"` here to open the link in a new tab. And also in other places with links (line 85 and 86 also)

src/components/Cards/CreateAccount.vue
```diff
@@ -65,12 +71,11 @@
                     @click.stop
                     v-on="on"
                   >
-                    Terms of Use
+                    Terms of Use.
                   </a>
                 </template>
                 Opens in new window
```
dati18, line 77: `Open link in new tab` (like Chrome) sounds better technically

src/components/Cards/CreateAccount.vue
rosalieper: the AC does not mention that. perhaps we should ask Charlie what she think about that.

src/components/Cards/CreateAccount.vue
rosalieper: same here, we should ask UX

src/components/Cards/CreateAccount.vue
rosalieper: As I understood Charlie, this is ok as is.

dati18 (approved):

tarrow:
Is the fix to the login button visible on the preview? Right now I see the button below the text whereas in figma it is beside (with some space).

image

vs

image

dati18:
Could you run it on you local env? Maybe the preview isn't updated

tarrow:
Create Account page looks good! I didn't dig into it in great detail but I think a similar thing is needed on the log-in page?
image

tarrow:
Login page button is now looking solid to me! Did you already checked the other design / layout features or does that still need to be done by someone?

PR #598, opened 2022-12-20
Adjust Log in page

dati18:


src/components/Pages/Login.vue
```diff
@@ -2,7 +2,7 @@
     <v-main>
       <v-container class="fill-height" fluid >
         <v-row align="center" justify="center">
-            <LoginCard title="Log in to your account" buttonText="Login"/>
+            <LoginCard title="Log in to your account" buttonText="Log in"/>
```
rosalieper, line 5: this should be "Log in" instead of "Log in to your account"

src/components/Cards/Login.vue
```diff
@@ -1,36 +1,36 @@
...
+          label="Password"
+          type="password"
+          required
+          v-model="password"
+          :disabled="loggingIn"
+          :error-messages="error['password']"
         />
       </v-form>
-      <router-link to="/forgotten-password">Forgotten password</router-link>
+      <router-link to="/forgotten-password" style="text-decoration: none">Forgot your password?</router-link>
```
tarrow, line 31: I wonder if we need to add the right (24px) padding here. Maybe using something like `py-6` added to the class list?

src/components/Cards/Login.vue
```diff
@@ -98,4 +98,8 @@ export default {
 </script>
 
 <style lang="css" scoped>
+.button-padding{
+  padding-right: 24px;
```
tarrow, line 102: I wonder if we should use the normal veutify padding helpers here (rather than tweaking the css ourselfs. 

Have a look at https://vuetifyjs.com/en/styles/spacing/#how-it-works

src/components/Cards/Login.vue
dati18: The padding bottom of the login field is 16px and the padding top of the login button is default to 8px. So luckily I set the right spacing to it accidentally.

rosalieper:
The icons on the login form are not the same as on the figma mock.

dati18:
> Overall, looks great. thank you. One thing i noticed, the icons on the login form are not the same as on the figma mock.

No worry, I've asked Charlie and she confirm the icons are ok

PR #601, opened 2022-12-23
Update docker.build.yml

rosalieper:
Open pull requests automatically for both stating and production,
after pushing to main in the UI component repository.

Bug: [T325814](https://phabricator.wikimedia.org/T325814)

### DIFF
--- a/.github/workflows/docker.build.yml
+++ b/.github/workflows/docker.build.yml
@@ -6,8 +6,6 @@ on:
       - 'main'
     tags:
       - '*'
-  pull_request:
-
 jobs:
   docker-build:
     runs-on: ubuntu-latest
@@ -57,3 +55,41 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
           labels: ${{ steps.docker_meta.outputs.labels }}
+      #Staging: Create pull request on deploy repo from a merged PR
+      -
+        name: create pull request for deploy on staging
+        uses: actions/checkout@v3
+        with:
+          repository: wmde/wbaas-deploy
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          branch: deploy-to-staging-${{ steps.docker_meta.outputs.tags }}
+          committer: wmdebot <github-bots@wikimedia.de>
+          author: ${{ github.actor }} <github-bots@wikimedia.de>
+          labels: automated pr
+          token: ${{ secrets.WBAAS_DEPLOY_TOKEN }}
+      #Production: Create pull request on deploy repo from a merged PR
+      - name: create pull request for deploy on production
+        uses: actions/checkout@v3
+        with:
+          repository: wmde/wbaas-deploy
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/production/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: 'Production: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          title: 'Production: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          branch: deploy-to-production-${{ steps.docker_meta.outputs.tags }}
+          committer: wmdebot <github-bots@wikimedia.de>
+          author: ${{ github.actor }} <github-bots@wikimedia.de>
+          labels: automated pr
+          token: ${{ secrets.WBAAS_DEPLOY_TOKEN }}


.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,19 @@ jobs:
...
-          labels: ${{ steps.docker_meta.outputs.labels }}
\ No newline at end of file
+          labels: ${{ steps.docker_meta.outputs.labels }}
+      #Create PR on deploy repo from a merged PR
+      -
+        name: create pull request for deploy on staging
+        uses: actions/checkout@v3
+        with:
+          repository: rosalieper/wbaas-deploy
+      # Make changes to pull request here
```
tarrow, line 66: This needs to be a separate step from the checkout one above

.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,19 @@ jobs:
...
+          run: |
+            yq -i '.image.tag = ${{ steps.docker_meta.outputs.tags }}' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+            yq -i '.image.tag = ${{ steps.docker_meta.outputs.tags }}' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
```
tarrow, line 70: I think there might need to be a `git add` and `git commit` step happening before this one

.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,19 @@ jobs:
           cache-to: type=local,mode=max,dest=/tmp/.buildx-cache
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.docker_meta.outputs.tags }}
-          labels: ${{ steps.docker_meta.outputs.labels }}
\ No newline at end of file
+          labels: ${{ steps.docker_meta.outputs.labels }}
+      #Create PR on deploy repo from a merged PR
+      -
+        name: create pull request for deploy on staging
```
tarrow, line 62: Looks like just running this (without specifying a path) is fine! That surprised me but it looks like this clears the current directory before checking it out (see: https://github.com/rosalieper/ui/actions/runs/3765335384/jobs/6400703190#step:9:28)

.github/workflows/docker.build.yml
tarrow: And before you can do that you need to initialise git I think (e.g. set the config). See something like: 
https://github.com/actions/checkout#push-a-commit-using-the-built-in-token

```
  build:
    runs-on: ubuntu-latest
    steps:
      - run: |
          # do some yq stuff here
          git config user.name github-actions
          git config user.email github-actions@github.com
          git add .
          git commit -m "some changes"
    # follow up with the create PR action
```

.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,19 @@ jobs:
...
+        uses: actions/checkout@v3
+        with:
+          repository: rosalieper/wbaas-deploy
+      # Make changes to pull request here
+          run: |
+            yq -i '.image.tag = ${{ steps.docker_meta.outputs.tags }}' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+            yq -i '.image.tag = ${{ steps.docker_meta.outputs.tags }}' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: Deploy new UI image
```
tarrow, line 72: What about trying to make the commit message + the title reflect the PR / commit that has been merged?

.github/workflows/docker.build.yml
tarrow: Or perhaps not? Perhaps this action below does actually do it all?

rosalieper: I think it does it all. The v4 api version allows that. I lost the link to the tutorial I followed but I will add it once i find it again

.github/workflows/docker.build.yml
rosalieper: you are right, thanks

.github/workflows/docker.build.yml
rosalieper: yep

m90 (commented):
I've left a couple of comments inline, most of them to gain understanding of how such an update is handled myself.

One thing I am wondering: considering this only updates local and staging, would it be an idea to create a 2nd PR that also updates production, so that it can be merged once we confirmed staging works as expected? Or do you usually batch updates to production and this is still better done manually?

---

For the record, I've built something very similar for an open source repo recently, this is what it looks like: https://github.com/offen/offen/blob/development/.github/workflows/heroku.yml

.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,24 @@ jobs:
...
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+        if: github.event.review.state == 'approved'
+      - uses: peter-evans/create-pull-request@v4
+        if: github.event.review.state == 'approved'
+        with:
+          commit-message: Staging+Local:Deploy new UI image ${{ steps.docker_meta.outputs.tags }}
+          title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          labels: automated pr
+          token: ${{ secrets.GH_TOKEN_PLUSW }}
```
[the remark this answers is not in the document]
rosalieper: Adam Shortland is the owner of the token and I think PLUSW is "plus write"

.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,24 @@ jobs:
...
+          repository: wmde/wbaas-deploy
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+        if: github.event.review.state == 'approved'
+      - uses: peter-evans/create-pull-request@v4
+        if: github.event.review.state == 'approved'
+        with:
```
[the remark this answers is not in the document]
rosalieper: I agree

.github/workflows/docker.build.yml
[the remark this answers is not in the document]
m90: If I understood the conversation in Mattermost correctly, we might want to use this user instead https://github.com/wmdebot who is managed by Leszek. This would make it easier to distinguish bot things from Adam things.

I think Conny (and all other EMs) should have the privileges required to create an access token for us. Should we ask her if she can help us?

.github/workflows/docker.build.yml
```diff
@@ -7,7 +7,7 @@ on:
     tags:
       - '*'
   pull_request:
-
+    types: ['opened', 'closed', 'reopened', 'synchronize', 'edited']
```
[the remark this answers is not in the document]
rosalieper: you are right. I think removing `edited` event would have it not build for every commit that's probably the only one we don't need i think. Perhaps, we want to keep the others. the default `pull request` `actions` that trigger the workflow are `opened, synchronize, reopened` it is not triggered by the `close` action by default https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

that was the reason i added that list  and the `===approved` because otherwise it creates the PR on `wbaas-deploy` even just from an open PR on UI.

.github/workflows/docker.build.yml
m90, line 75: We will also need to set a unique `branch` value, else we'll get conflicts if two PRs are opened by this at the same time.

.github/workflows/docker.build.yml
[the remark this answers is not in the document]
m90: Ah ok, I didn't know that `pull_request:` itself already sets up triggers, I was thinking this meant "do nothing on a PR".

For my understanding: why do we need to build and publish images for PRs? Is this just to see if the image builds fine?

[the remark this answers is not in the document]
rosalieper: I also would prefer another way to do this as having `===approved` line written 3 times in the same file is not ideal.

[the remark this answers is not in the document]
m90: > I also would prefer another way to do this as having ===approved line written 3 times in the same file is not ideal.

I'm not entirely sure if we should even open an automated PR on approval because of the following scneario:

- I open a PR to this repo with some changes
- There's some back and forth with changes, I push a few commits
- The PR gets approved
- Now, this action opens a PR against `wbaas-deploy` using the git SHA from my feature branch
- I merge my PR and all of my commits get squashed into `main` creating a new commit. Ideally, I delete my feature branch.
- The SHA that we now want to deploy in `wbaas-deploy` does not exist in the repo anymore as:
  - `main` has different commits
  - my feature branch was deleted

Obviously, this works as the image is on `ghcr` already, but it's probably pretty confusing when you want to check which code is currently running in the cluster?

To me, the easiest way of solving this would having this action run on changes to `main` only, but I might be missing something you used to do with the images built off PR branches. If we just want to check that the image builds, we can also add a second action that does just that for PRs and not for `main`.

.github/workflows/docker.build.yml
```diff
@@ -56,4 +56,24 @@ jobs:
...
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+        if: github.event.review.state == 'approved'
+      - uses: peter-evans/create-pull-request@v4
+        if: github.event.review.state == 'approved'
+        with:
+          commit-message: Staging+Local:Deploy new UI image ${{ steps.docker_meta.outputs.tags }}
```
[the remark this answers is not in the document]
rosalieper: good catch, I fixed it

.github/workflows/docker.build.yml
rosalieper: I added a branch

.github/workflows/docker.build.yml
```diff
@@ -56,4 +54,43 @@ jobs:
...
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: Staging+Local:Deploy new UI image ${{ steps.docker_meta.outputs.tags }}
+          title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          labels: automated pr
+          token: ${{ secrets.WBAAS_DEPLOY_TOKEN }}
+      #Production: Create pull request on deploy repo from a merged PR
+      -
+        name: create pull request for deploy on production
+        uses: actions/checkout@v3
```
m90, line 81: The repository has already been checked out in one of the steps above. Is it safe/expected to run the action again in order to reset the repo state to current `main`?

.github/workflows/docker.build.yml
```diff
@@ -56,4 +54,43 @@ jobs:
...
+          committer: wmdebot <github-bots@wikimedia.de>
+          author: ${{ github.actor }} <github-bots@wikimedia.de>
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: Staging+Local:Deploy new UI image ${{ steps.docker_meta.outputs.tags }}
```
m90, line 74:
```suggestion
          commit-message: Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}
```

.github/workflows/docker.build.yml
```diff
@@ -56,4 +54,43 @@ jobs:
...
+          committer: wmdebot <github-bots@wikimedia.de>
+          author: ${{ github.actor }} <github-bots@wikimedia.de>
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/production/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: 'Production: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
```
m90, line 94:
```suggestion
          title: 'Production: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
```

.github/workflows/docker.build.yml
```diff
@@ -56,4 +54,43 @@ jobs:
...
+          labels: automated pr
+          token: ${{ secrets.WBAAS_DEPLOY_TOKEN }}
+      #Production: Create pull request on deploy repo from a merged PR
+      -
+        name: create pull request for deploy on production
+        uses: actions/checkout@v3
+        with:
+          repository: wmde/wbaas-deploy
+          branch: deploy-to-production-${{ steps.docker_meta.outputs.tags }}
```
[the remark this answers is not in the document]
rosalieper: you are right

.github/workflows/docker.build.yml
m90: Looks like it is safe: https://github.com/actions/checkout/issues/33

.github/workflows/docker.build.yml
```diff
@@ -56,4 +54,42 @@ jobs:
...
+      # Make changes to pull request here
+      - run: |
+          get_tag=$(echo ${{ steps.docker_meta.outputs.tags }} | cut -d':' -f2)
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          yq -i '.image.tag=$get_tag' k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+      - uses: peter-evans/create-pull-request@v4
+        with:
+          commit-message: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
+          branch: deploy-to-staging-${{ steps.docker_meta.outputs.tags }}
```
m90, line 73: Seems this breaks because the branch name would now contain colons, which is not allowed.

We probably can use `deploy-to-staging-${get_tag}` instead which should be unique as well (we'd need to make it available across steps or use something similar though).

Sorry for not catching this in my initial review.

PR #621, opened 2023-01-18
change branch name for created PRs in workflow

rosalieper:


.github/workflows/docker.build.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         with:
           commit-message: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
           title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
-          branch: deploy-to-staging-${{ steps.docker_meta.outputs.tags }}
+          branch: deploy-to-staging
```
m90, line 73: I think these need to be unique still, else we'll create conflicts in the `wbaas-deploy` repo when people do not instantly merge these PRs. Maybe we can find something else that is unique, but doesn't contain colons (which is why this broke). Maybe we can use `GITHUB_RUN_ID` instead which is a unique identifier for each workflow run: https://docs.github.com/en/actions/learn-github-actions/variables

rosalieper: that is a great idea. thanks

.github/workflows/docker.build.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         with:
           commit-message: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
           title: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
-          branch: deploy-to-staging-${{ steps.docker_meta.outputs.tags }}
+          branch: deploy-to-staging-${{GITHUB_RUN_ID}}
```
m90, line 73: I'm not entirely sure how to access env vars in action parameters. I can't find anything in the official docs right now, but random examples from Googling are all using `${{ env.GITHUB_RUN_ID }}`

rosalieper: ok

m90: Ah, yeah looks like this should work https://docs.github.com/en/actions/learn-github-actions/contexts#example-usage-of-the-env-context

m90 (approved):
Let's give this another try :crossed_fingers:

PR #622, opened 2023-01-18
Make the `create pull request job` in the work flow not to run

rosalieper:


.github/workflows/docker.build.yml
```diff
@@ -59,14 +59,20 @@ jobs:
       -
         name: create pull request for deploy on staging
         uses: actions/checkout@v3
+        # remove this if statement after fixing the PR
+        if: 1==2
```
m90, line 63: This could also be just `false` I guess, but I like the condition just because we can.

m90 (approved):
Let's see if we can get this working on the next attempt.

PR #623, opened 2023-01-18
WIP: Fix the yaml edit

rosalieper:


.github/workflows/docker.build.yml
```diff
@@ -56,23 +56,36 @@ jobs:
...
+      - run:
+          |
+          python - <<EOF
+          from ruamel.yaml import YAML
+          import sys
+
+          yaml=YAML()
+          with open("k8s/helmfile/env/local/ui.values.yaml.gotmpl") as doc:
+              data=yaml.load(doc)
+          data['image']['tag']="sha-"+${{env.GITHUB_SHA}}[:7]
```
tarrow, line 79: looks like perhaps this `${{env.GITHUB_SHA}}[:7]` isn't quite doing it; I wonder if that's because this isn't being correctly parsed or instead if this environment variable is just empty

rosalieper: The variable is empty. It has not value `on pull_request` since this branch has no merge commits but has one after a push and it is the latest commit in main. see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

On push see: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push

tarrow:
Looks like there might still be a little work to not break the null. Maybe this can help? https://stackoverflow.com/a/57207057

Also it would be good to fix the branch name; right now it's literally (e.g.) `deploy-to-production-` without any suffix. This means that if more than one PR is open at a timer there will be collisions.

PR #626, opened 2023-01-23
Fix updating of image tag versions

m90:
Supersedes #623 

Doing this on a new branch so I don't eat potential local changes from @rosalieper  

Example PRs opened by this approach:
- https://github.com/wmde/wbaas-deploy/pull/703
- https://github.com/wmde/wbaas-deploy/pull/702


.github/workflows/docker.build.yml
```diff
@@ -3,9 +3,7 @@ name: Docker build
 on:
   push:
     branches:
-      - 'main'
-    tags:
```
m90, line 7: I deleted this by accident, but I'm not sure if this is even needed in the current workflow?

.github/workflows/docker.build.yml
```diff
@@ -52,56 +49,54 @@ jobs:
...
+          echo "tag=$TAG" >> $GITHUB_OUTPUT
+          sed -i "/image:/{n;s/tag:.*/tag: $TAG/;}" ./repos/wbaas-deploy-staging/k8s/helmfile/env/local/ui.values.yaml.gotmpl
+          sed -i "/image:/{n;s/tag:.*/tag: $TAG/;}" ./repos/wbaas-deploy-staging/k8s/helmfile/env/staging/ui.values.yaml.gotmpl
+          sed -i "/image:/{n;s/tag:.*/tag: $TAG/;}" ./repos/wbaas-deploy-production/k8s/helmfile/env/production/ui.values.yaml.gotmpl
       - uses: peter-evans/create-pull-request@v4
-        # TODO: Remove this if statement after fixing the PR
-        if: 1==2
         with:
+          path: ./repos/wbaas-deploy-staging
           commit-message: 'Staging+Local: Deploy new UI image ${{ steps.docker_meta.outputs.tags }}'
```
m90, line 78: This is a very theoretical concern, but I wonder if we want to replace the "deploy" terminology in these PRs with "update" as no automated deployment is happening, it's more of an automated update a human would then deploy.

.github/workflows/docker.build.yml
tarrow: Yeah, we probably are now safely away from the "bothering to tag" workflow

.github/workflows/docker.build.yml
tarrow: Yeah, I can see us looking at making this more meaningful as time goes by. Right now I think it will do

tarrow (approved):
Looks good to me; lets try this thing for a few days :)

PR #640, opened 2023-02-03
Add new logo to mainpage

dati18:


src/components/Layout/Navbar.vue
```diff
@@ -1,9 +1,8 @@
 <template>
   <v-toolbar>
-    <v-toolbar-title>
-      <!-- TODO remove link styling...-->
-      <router-link to="/">wikibase.cloud</router-link>
-    </v-toolbar-title>
+    <a class="pr-4" href="/" style="height: 64px">
+      <img src="../../assets/logo.svg" style="height: 64px">
```
rosalieper, line 4: I am not a big fan of inline css and prefer to avoid if possible. But I think it all looks good

rosalieper (approved):

src/components/Layout/Navbar.vue
dati18: you mean the helper class `pr-4`?
Anyway the `v-toolbar` has the default height of 64px but if you zoom in or open it on mobile it will automatically change to 56px so the logo will look a bit off center. Is it worth patching up?

rosalieper: what I meant by inline css is `style=height: 64px`

dati18: @rosalieper I've changed it to internal css :D looks cleaner now

PR #648, opened 2023-02-21
Fix Login error feedback handling

deer-wmde:
This aims to fix the issue that the Login form does "freeze" in a gray state if the login fails and the API does not provide additional details.

### DIFF
--- a/src/components/Cards/Login.vue
+++ b/src/components/Cards/Login.vue
@@ -78,10 +78,7 @@ export default {
           console.log(err.response)
 
           // If the api gave use details of the error, then use them
-          if (err.response !== undefined &&
-                err.response.data !== undefined &&
-                err.response.data.errors !== undefined
-          ) {
+          if (err.response && err.response.data && err.response.data.errors) {
             if (err.response.data.errors.email) {
               this.error.email = err.response.data.errors.email[0]
             }
--- a/src/components/Cards/Login.vue
+++ b/src/components/Cards/Login.vue
@@ -78,7 +78,10 @@ export default {
           console.log(err.response)
 
           // If the api gave use details of the error, then use them
-          if (err.response.data.errors) {
+          if (err.response !== undefined &&
+                err.response.data !== undefined &&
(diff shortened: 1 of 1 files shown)


src/components/Cards/Login.vue
```diff
@@ -78,7 +78,10 @@ export default {
           console.log(err.response)
 
           // If the api gave use details of the error, then use them
-          if (err.response.data.errors) {
+          if (err.response !== undefined &&
```
deer-wmde, line 81: Does someone know a shorthand for achieving the same thing? I'm sure it can be simplified somehow

src/components/Cards/Login.vue
```diff
@@ -78,7 +78,10 @@ export default {
...
+                err.response.data !== undefined &&
+                err.response.data.errors !== undefined
+          ) {
```
deer-wmde, line 84: Does someone know a shorthand for achieving the same thing? I'm sure this `if` statement can be simplified somehow

m90: The only thing I can think of (which would also shield this from `null`s which currently still would trip up the condition) is checking for truthiness instead:

```js
if (err.response && err.response.data && err.response.data.errors) {
```
It doesn't look like we've got Lodash or similar floating around in which case you could `_.get(err, 'response.data.errors')`.

Another option would be replacing the if/else with a try/catch that starts with
```js
const { errors } = err.response.data
```

The first option or exactly what you already have written might be our best bet.

deer-wmde: I like the first option, thanks!

m90 (approved):

src/components/Cards/Login.vue
deer-wmde: I was just cleaning some tabs and will leave this here for future-me: I think optional chaining could be used in such cases https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

PR #651, opened 2023-03-06
fix(ci): netlify preview does not always delegate routing to SPA

m90:
In a Netlify preview it's currently not possible to "browser reload" a non-index page, e.g. `/login` (a 404 page is shown instead). This PR makes Netlify delegate routing of requests that do not match a file to Vue's SPA router, allowing users to reload any page while testing.


tarrow:
Looks fine to me; for people wondering how the magic works look at: https://docs.netlify.com/routing/redirects/

tarrow (approved):

PR #660, opened 2023-05-10
Make navbar responsive for small screen sizes

rosalieper:
Bug: T334221


src/components/Layout/Navbar.vue
```diff
@@ -11,7 +11,20 @@
...
+          </v-btn>
+        </template>
+        <v-list>
+          <v-list-item>
+            <v-btn id="nav-login" text to="/login">Log in</v-btn>
+            <v-btn id="nav-create-account" ext color="primary" to="/create-account">Sign up</v-btn>
+          </v-list-item>
+        </v-list>
+        </v-menu>
+      <template v-else>
```
m90, line 27: Issue: does this also need to consider the log in status, i.e. be `else if !isLoggedIn`?

src/components/Layout/Navbar.vue
```diff
@@ -11,7 +11,20 @@
...
-      <template v-if="!isLoggedIn">
+      <v-menu bottom left v-if="$vuetify.breakpoint.xs && !isLoggedIn">
+        <template v-slot:activator="{ on, attrs }">
+          <v-btn icon v-bind="attrs" v-on="on">
+              <v-icon>menu</v-icon>
+          </v-btn>
+        </template>
+        <v-list>
+          <v-list-item>
+            <v-btn id="nav-login" text to="/login">Log in</v-btn>
```
m90, line 22: Question: Is there an easy way to deduplicate this and use the same elements as defined below?

m90 (commented):
This seems to work as intended (one issue to fix still I think), although it's a bit unfortunate the ticket does not specify anything about how the buttons should look like when expanded (I would have expected sth hamburger-ish).

m90:
Ah I found the Figma design for this. Seems the buttons are expected to look different than in this version?

image

Also, a different icon is used when in collapsed state:

image

src/components/Layout/Navbar.vue
```diff
@@ -11,9 +11,28 @@
         <v-btn id="nav-user" text to="/user">Account</v-btn>
         <v-btn id="nav-logout" text to="/logout">Logout</v-btn>
       </template>
-      <template v-if="!isLoggedIn">
-        <v-btn id="nav-login" text to="/login">Log in</v-btn>
-        <v-btn id="nav-create-account" ext color="primary" to="/create-account">Sign up</v-btn>
+      <template v-else-if="$vuetify.breakpoint.xs && !isLoggedIn">
```
m90, line 14: My bad for not catching this in the last review where I did not see that there is a `template` block above.

Would it be possible to make the conditions simpler by using:
```
<template v-if="isLoggedIn` />
<template v-else-if="$vuetify.breakpoint.xs" />
<template v-else />
```

Maybe that's a bit easier to understand?

m90 (approved):
Looks good to me as a code change now :tophat: . Let's have Charlie review the visual still.

src/components/Layout/Navbar.vue
```diff
@@ -11,9 +11,26 @@
         <v-btn id="nav-user" text to="/user">Account</v-btn>
         <v-btn id="nav-logout" text to="/logout">Logout</v-btn>
       </template>
-      <template v-if="!isLoggedIn">
-        <v-btn id="nav-login" text to="/login">Log in</v-btn>
-        <v-btn id="nav-create-account" ext color="primary" to="/create-account">Sign up</v-btn>
+      <v-menu bottom left v-if="($vuetify.breakpoint.width < 400)  && !isLoggedIn">
+        <template v-slot:activator="{ on, attrs }">
+          <v-btn class="dots-button" v-bind="attrs" v-on="on">
```
dati18, line 16: Neat Picking: I added `.dot-button` to remove the `box-shadow` which was turned on by default. It also changed the `background-color` of the button to `transparent`, it was #f5f5f5 by default, which looked a bit off.

src/components/Layout/Navbar.vue
```diff
@@ -11,9 +11,26 @@
         <v-btn id="nav-user" text to="/user">Account</v-btn>
         <v-btn id="nav-logout" text to="/logout">Logout</v-btn>
       </template>
-      <template v-if="!isLoggedIn">
-        <v-btn id="nav-login" text to="/login">Log in</v-btn>
-        <v-btn id="nav-create-account" ext color="primary" to="/create-account">Sign up</v-btn>
+      <v-menu bottom left v-if="($vuetify.breakpoint.width < 434)  && !isLoggedIn">
```
dati18, line 14: I check figma design and Charlie set the breakpoint (the point where login/logout buttons merge into dots button) at 434px. I want to consult with Charlie to up the breakpoint to =<576px which cover larger number of mobile devices and 576px is also the default breakpoint of VueJS.

dati18: But I'm happy as it is now. I also check some JS framework and their `small` breakpoints are around 460-600px

m90 (changes requested):
When expanded, the current menu

image

does not match the menu in Figma

image

I.e. casing is different as well as the backdrop on the buttons.

m90 (dismissed):

dati18:
Screenshot from 2023-06-21 17-08-10

Anyway the drop-down menu is too small for the text in your recent patch

dati18:
some [thoughts](https://phabricator.wikimedia.org/T334221#9046488:~:text=At%20the%20moment%2C%20the%20dropdown%20menu%20overlaps%20the%203%2Ddots%2Dicon%20and%20it%27s%20default%20by%20Vuetify%20v2%2C%20which%20is%20the%20version%20we%20are%20using.%20In%20Vuetify%20you%20can%27t%20use%20style%20or%20class%20to%20set%20custom%20style%20for%20%3Cv%2Dmenu%3E%20so%20the%20dropdown%20menu%20stuck%20at%20that%20position.) about the ticket

deer-wmde:
I just noticed that in the current state, if logged in
1. all of the buttons are being displayed (Dashboard, Account, Logout, Log In, Sign Up)
2. we basically need the same fix for the buttons that are (regularly) displayed when logged in

dati18:
> I just noticed that in the current state, if logged in
> 
> 1. all of the buttons are being displayed (Dashboard, Account, Logout, Log In, Sign Up)
> 2. we basically need the same fix for the buttons that are (regularly) displayed when logged in

@deer-wmde I removed SignUp and LogIn button when logged in

deer-wmde (approved):

PR #714, opened 2023-08-14
fix(workflow): quoting in truncate-commit-message

deer-wmde:
Double quotes in commit messages broke the `truncate-commit-message` workflow step of `docker-build` - this hopefully fixes it

example: https://github.com/wbstack/ui/actions/runs/5853401197/job/15867755875#step:12:21

### DIFF
--- a/.github/workflows/docker.build.yml
+++ b/.github/workflows/docker.build.yml
@@ -77,7 +77,7 @@ jobs:
         name: Truncate commit message
         id: truncate-commit-message
         run: |
-          MSG=$(echo "${{ github.event.head_commit.message }}" | head -n 1)
+          MSG=$(echo '${{ github.event.head_commit.message }}' | head -n 1)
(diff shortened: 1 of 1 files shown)


deer-wmde:
> Not sure I entirely understand the error (there should be a closing quote, no?), but let's try this.

I think it is because there is no escaping going on, the string `"Revert "Bump node-sass from 6.0.1 to 9.0.0 (#666)"` already terminates after `Revert ` and the rest is then basically unquoted. I think the `()` then throws bash off

deer-wmde:
Seems to work https://github.com/wbstack/ui/actions/runs/5853746969/job/15868344747#step:12:1

deer-wmde:
> Seems to work https://github.com/wbstack/ui/actions/runs/5853746969/job/15868344747#step:12:1

Actually I think it probably still breaks if one uses single quotes ...

PR #725, opened 2023-09-06
Run node.js checks only for our current version

AndrewKostka:


tarrow (commented):
Might be better to keep 15 and add 16 / 18? If either works we should then move to that?

PR #759, opened 2023-12-04
Redesign of the homepage

AndrewKostka:


deer-wmde (approved):
technical approval from me - we can iterate on the implementation later as discussed in refinement :shipit:

PR #762, opened 2024-01-19
Temporarily disable custom logo feature

AndrewKostka:


tarrow (approved):
Looks good to me; let's ship this as a when when maybe break the feature for a bit

m90:
Closing this as there is no need to disable the feature. Sorry for having caused some confusion apparently.

PR #766, opened 2024-02-05
Fix password reset page icons

addshore:
Bug: T355665


deer-wmde:
good catch, thanks :)

PR #776, opened 2024-02-08
fix: lexemes cannot be disabled as button is rendered outside of form

m90:
Ticket https://phabricator.wikimedia.org/T353533

### DIFF
--- a/src/components/Pages/ManageWiki/Features/Lexeme.vue
+++ b/src/components/Pages/ManageWiki/Features/Lexeme.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-form ref="form" v-model="valid" @submit="confirmAction">
+  <v-form ref="form" v-model="valid" @submit="handleSubmit" id="formLexeme">
     <v-card elevation="1" class="padding" outlined>
       <v-card-title>Lexeme</v-card-title>
       <v-card-text>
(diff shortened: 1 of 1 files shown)


src/components/Pages/ManageWiki/Features/Lexeme.vue
```diff
@@ -53,41 +54,47 @@ export default {
...
   props: ['wikiId', 'setting'],
   data () {
-    return {
+    const data = {
       valid: false,
       dialog: false,
       userConfirmationString: 'wbstack-important-change',
-      rules: {
-        match: (value) =>
-          value === this.userConfirmationString || 'Please enter the confirmation'
```
m90, line 62: Second problem was that `this` here would be the Vue component, not the `data` object, thus returning `undefined` instead of the value of `userConfirmationString`

Also the comparison is borked and would return either `true|'Please enter the confirmation'` instead of `true|false`

src/components/Pages/ManageWiki/Features/Lexeme.vue
```diff
@@ -28,15 +28,16 @@
...
-        <v-divider></v-divider>
+        <v-divider>
+        }</v-divider>
 
         <v-card-actions>
           <v-spacer></v-spacer>
           <v-tooltip top>
             <template v-slot:activator="{ on, attrs }">
-              <v-btn color="primary" v-bind="attrs" v-on="on" text type="submit">
+              <v-btn v-bind="attrs" v-on="on" color="primary" type="submit" form="formLexeme">
```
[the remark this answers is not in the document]
tarrow: I think this looks nice

src/components/Pages/ManageWiki/Features/Lexeme.vue
tarrow: I'm very unclear how this ever really worked

tarrow (approved):
lgtm; also checked it worked as expected in the prod version of netlify

PR #786, opened 2024-03-06
Remove unnecessary columns from all pages and from footer

rosalieper:
Bug: T350370

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -3,11 +3,7 @@
       <Navbar></Navbar>
       <router-view v-if="customLayout"/>
       <v-container v-else class="full-height-content">
-        <v-row >
-          <v-col>
-            <router-view></router-view>
-          </v-col>
-        </v-row>
+        <router-view></router-view>
       </v-container>
       <Foot :class="{'tall-footer': customLayout}"></Foot>
       <Interval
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -4,11 +4,9 @@
       <router-view v-if="customLayout"/>
       <v-container v-else class="full-height-content">
         <v-row >
-          <v-col cols="1"></v-col>
-          <v-col cols="10">
+          <v-col>
```
m90, line 7: Idea: if we now have a row that contains a single column, there is no need for a grid system anymore, and we could also just remove both the `v-row` and the `v-col` entirely with the `v-container` making sure everything stays in place.

m90 (commented):
This works as described in the ticket. I left a suggestion about how to further simplify this, lmk what you think.

src/App.vue
rosalieper: I am leaving container in this case because of the class.

src/components/Layout/Foot.vue
```diff
@@ -1,10 +1,6 @@
 <template>
   <footer class="footer">
   <v-footer color="primary lighten-1">
-    <v-container>
```
m90, line 4: I think the container is still needed for giving the footer content a width, it's just `col` and `row` that are superfluous.

image

m90 (approved):
:sparkles:

PR #791, opened 2024-03-13
nginx: add Cache-Control header

deer-wmde:
https://phabricator.wikimedia.org/T331423

Adds the `Cache-Control` header to the nginx response. Forces revalidation for `index.html`

### DIFF
--- a/nginx/default.conf
+++ b/nginx/default.conf
@@ -21,6 +21,8 @@ server {
   location / {
       root /usr/share/nginx/html;
       try_files $uri /index.html;
+
+      add_header 'Cache-Control' 'max-age: 31536000, immutable' always;
   }
 
   location = /index.html {
--- a/nginx/default.conf
+++ b/nginx/default.conf
@@ -22,9 +22,11 @@ server {
       root /usr/share/nginx/html;
       try_files $uri /index.html;
 
-      add_header 'Cache-Control' 'max-age: 31536000, immutable' always;
+      # cache everything for 1 week
+      add_header 'Cache-Control' 'max-age: 604800, immutable' always; 
   }
 
+  # do not cache index.html - https://phabricator.wikimedia.org/T331423
   location = /index.html {
       add_header 'Cache-Control' 'no-cache' always;
   }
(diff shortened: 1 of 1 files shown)


m90:
I would think we still need to set different values for assets that are not `index.html`, i.e. revisioned JS and CSS assets that will never change and thus can be cached:

> JS and CSS bundles are revisioned by the build system. However, we do not send a Cache-Control header here either. As these files will never change after being created (as opposed to index.html), we can set "hard" cache control values to make sure users will never have to redownload the same bundle twice. Something like Cache-Control: max-age: 31536000, immutable should work well.

deer-wmde:
I wonder if we should lower the `max-age` for everything else though. I'm thinking of assets like images that may not change their filename (I think?)

m90:
> I'm thinking of assets like images that may not change their filename (I think?)

A cursory check tells me all assets that are handled by the Vue App are revisioned. We do load some stuff from GCS as well, but those aren't affected by this change anyways.

nginx/default.conf
```diff
@@ -21,6 +21,14 @@ server {
   location / {
       root /usr/share/nginx/html;
       try_files $uri /index.html;
+
+      # cache everything for 1 week
+      add_header 'Cache-Control' 'max-age: 604800, immutable' always;
```
m90, line 26: Considering all assets should be revisioned, I think a year or even longer would be fine as well. In case we do serve non-revved assets through this setu, we might want to check why?

deer-wmde:
> > I'm thinking of assets like images that may not change their filename (I think?)
> 
> A cursory check tells me all assets that are handled by the Vue App are revisioned. We do load some stuff from GCS as well, but those aren't affected by this change anyways.

Indeed! I didn't know this, amazing. In that case we should be good, thanks for mentioning this

m90:
> Indeed! I didn't know this, amazing. In that case we should be good, thanks for mentioning this

Unfortunately I was wrong and we're serving an unrevisioned `/config.js` file which we'll need to handle just like `index.html` I would assume.

m90 (approved):
Works well for me when testing this locally.

On subsequent loads, `index.html` and `config.js` will be requested from the server, yielding a 304 response, all other assets will skip the request, reading from disk cache.

When adding arbitrary changes, I will the correct version after reload.

PR #801, opened 2024-03-25
T347865: Ask for feedback before deleting a wikibase instance.

rosalieper:


src/components/Pages/ManageWiki/Cards/ConfirmDelete.vue
```diff
@@ -0,0 +1,72 @@
...
+    <v-card>
+      <v-card-title>
+        Confirm Deletion
+      </v-card-title>
+      <v-card-text class="align-start pb-4">
+        Before you delete your Wikibase instance, <br/>please let us know the reason for your deletion.<br/>
+        Please select all that apply.
+      </v-card-text>
+      <div class="px-4 pb-2">
+        <v-checkbox class="ma-0" hide-details label="Was only used for Testing" @click="deletion_reason='Was only used for testing'"/>
```
deer-wmde, line 12: I noticed that by using `@click` here like this, the `deletion_reason` gets set every time you click it, not only when you check it. So unchecking a box actually sets it to that reason string currently. Also, this way only 1 reason is getting passed to the API (namely the last one clicked). 

So my suggestions would be to
1. check if the click actually checked the checkbox
  - nitpick: also `@change` might be better suited here instead of `@click`, but I'm not sure if it plays nicely with `v-checkbox`
2. collect all selected reasons instead of just the last one
  - for example `deletion_reason` could be an object, which you could transmit to the API as a JSON string, something like `{ justTesting: false, tooComplex: true, ..., other:"" }`

deer-wmde:
yay! this looks great now as far as I can tell :partying_face:

PR #817, opened 2024-05-06
Rely on authentication token stored in cookie

m90:
Ticket https://phabricator.wikimedia.org/T363100


nginx/default.template.conf
```diff
@@ -17,6 +17,10 @@ server {
   root /usr/share/nginx/html;
   index index.html index.html;
 
+  location /api/(.*) {
+    proxy_pass ${API_URL}/$1;
```
tarrow, line 21: suggestion: in a follow up PR we could add some docs to make it clear that this env var is now to be defined as the URL from the perspective of the server hosting the UI and not the client accessing it

tarrow (approved):
looks good to me; nothing extra to add after our long chat last week

PR #820, opened 2024-06-03
Mainpage e2e test

dati18:


src/components/Layout/Foot.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <footer class="footer">
+  <footer id="footer" class="footer">
```
m90, line 2: Suggestion: A good pattern I've seen used elsewhere is using `data-test-id` attributes instead of `id` proper which reduces the risk of inadvertent side effects (this now creates `window.footer`) and also makes it very clear what the attribute is about, i.e.

```html
<footer data-test-id="footer" class="footer">
```

and then lateron you select the element using `$('[data-test-id="footer"]')`

This comment also applies to all other places where `id`s are introduced.

docker-compose.yml
```diff
@@ -8,6 +8,22 @@ services:
     working_dir: /home/node/app
     volumes:
     - ./:/home/node/app
-    command: npm run serve -- --port 8081
+    command: npm run serve -- --port 8081 --host ui
     ports:
       - 8081:8081
+    healthcheck:
+      test: curl --fail http://ui:8081 || exit 1
```
m90, line 15: Question: Is `|| exit 1` needed when `curl --fail` is used? I would think this already exits non-zero on failure, so there's no need to repeat this.

.github/workflows/browser-tests.js.yml
```diff
@@ -0,0 +1,18 @@
...
+
+jobs:
+  browser-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4.1.6
+      - run: cp .env.development.prod .env
+      - run: sudo chown -R 1000:1000 ${GITHUB_WORKSPACE}
+      - run: docker compose run ui npm install
```
m90, line 16:
```suggestion
      - run: docker compose run --rm ui npm install
```
No need to keep the dangling container post-install.

tests/e2e/specs/app.spec.js
```diff
@@ -1,13 +1,96 @@
...
+    const navCreateAccount = await $('#nav-create-account')
+    // await navCreateAccount.waitForDisplayed()
+
+    // Assert that the element is clickable
+    expect(await navCreateAccount.isClickable()).toBe(true)
+    await navCreateAccount.click()
+  })
+
+  it('should collapse Login and Signup buttons into a dots-button icon when screen width < 600px', async () => {
+    await browser.setWindowSize(599, 800)
```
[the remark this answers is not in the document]
m90: +1 on this, which should still be adressed.

tests/e2e/specs/app.spec.js
```diff
@@ -1,13 +1,96 @@
...
+  it('should render the WBC logo', async () => {
     App.open()
-    expect(App.navLogin).toExist()
-    expect(App.navCreateAccount).toExist()
+    expect(App.logo).toExist()
   })
-  it('should open and render with some buttons', () => {
+
+  it('should open and render with Login button', async () => {
+    const navLogin = await $('#nav-login')
```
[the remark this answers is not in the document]
m90: +1 on this, if we store some selectors on the page object class, let's store all of them there.

.github/workflows/browser-tests.js.yml
```diff
@@ -0,0 +1,18 @@
+name: Browser tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  browser-tests:
+    runs-on: ubuntu-latest
```
m90, line 10: Question: should we pin the ubuntu version here, so we don't get surprise updates at some point in the future?

dati18: which version you suggested? Or I just use the current one because it works?

m90: It seems `24.04` is the latest one available https://github.com/actions/runner-images

src/components/Layout/Foot.vue
dati18: good idea. I also think that it could be troublesome...

tests/e2e/specs/app.spec.js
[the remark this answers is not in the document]
dati18: I added this (line 4-7)

src/components/Layout/Navbar.vue
```diff
@@ -28,7 +28,7 @@
       </v-menu>
       <template v-if="($vuetify.breakpoint.width >= 600) && !isLoggedIn">
         <v-btn
-          id="nav-login"
+          data-test-id="nav-login"
```
m90, line 31: While it's good to use the `data-test-id`s for testing, I'm not sure it's safe to remove existing `id` attributes. We'd better leave them in here so code that expects them to exist (in case it exists?) keeps finding them.

m90 (approved):

PR #826, opened 2024-07-22
Add Data Reuse functionality

m90:
Ticket https://phabricator.wikimedia.org/T368026

### DIFF
--- a/src/backend/api.js
+++ b/src/backend/api.js
@@ -87,3 +87,21 @@ export const wikiDiscovery = async ({ sort, direction, active, currentPage, resu
     }
   })).data
 }
+
+export const importEntities = async ({
+  wikiId,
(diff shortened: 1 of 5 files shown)


m90:
This is very hard to test locally, so I would advise using the mock API and tampering with the responses there when testing.

src/backend/api.js
```diff
@@ -87,3 +87,21 @@ export const wikiDiscovery = async ({ sort, direction, active, currentPage, resu
     }
   })).data
 }
+
+export const importEntities = async ({
+  wikiId,
+  entityIds = ['P22', 'P25', 'P40', 'P18', 'P31', 'P279', 'Q1', 'Q2', 'Q64', 'Q42', 'Q3107329'],
```
m90, line 93: Not entirely sure if this function's default args are the right place to put our preset's configuration for now?

PR #828, opened 2024-07-31
chore: lock entity ids to pre-mul revisions

m90:
Ticket https://phabricator.wikimedia.org/T360031

We found the recent introduction of the `mul` language code in wikidata led to mass deletion of redundant labels on certain entities already. These entities would now have sparsely populated labels when imported into cloud, which is lacking `mul` support.

The revisions were picked as "either latest of today in case no `mul` data is present, or the last revision without `mul` data". (Only Q42 has mul data as of now).


tarrow (approved):
lgtm; didn't manually check each revision is exactly what we want

PR #848, opened 2024-08-13
Update to Node.js 22

AndrewKostka:


dati18 (approved):

src/backend/api.js
```diff
@@ -8,15 +8,8 @@ export const login = async (user) => {
   } else {
     call = axios.get('/auth/login')
   }
-  try {
-    const { data } = await call
-    return data
-  } catch (err) {
-    if (err.response.code === '401') {
```
m90, line 15: It's not safe to remove the 401 check which was deliberately introduced here https://github.com/wbstack/ui/commit/00120ef2243e70aab8f9a30eab02a03d6899a689

When loading the application, a call against login is made to check whether the user is already logged in. If a 401 is returned, this means the user is not, but it's an expected outcome, so `null` is being returned instead of throwing an error.

src/store/auth.js
```diff
@@ -48,7 +48,9 @@ const actions = {
       })
       .catch((error) => {
         commit('auth_error')
-        throw error
+        if (error.response.status !== 401) {
```
m90, line 51: I'm not sure this is equivalent to what happens now as `auth_error` is being comitted when previously it was not. API returning 401 is expected and not to be considered an error.

m90 (commented):
I'm not sure the login HTTP call change is something we should be doing. Is there are particular reason this is now needed as part of the upgrade?

.github/workflows/node.js.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12, 14, 15]
+        node-version: [22]
```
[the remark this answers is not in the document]
tarrow: We probably want to always test the *current* version and then perhaps any future versions we hope to update to soon

dati18 (approved):
since Node 22 is the newest and it runs fine. Let's merge it

PR #852, opened 2024-08-14
Run docker build on PR

dati18:


m90:
Over in `api` we also build the image on Pull Requests, but don't push them, I guess we want the same here: https://github.com/wbstack/api/blob/184cdd4f063a9ed70ac948be8ce6c1d79e11e420/.github/workflows/docker.build.yml#L55

dati18:
@m90 please check again

m90:
I think the additional trigger is still needed. See the action definition for API as a reference

dati18:
can you elaborate @m90 ?

m90:
Right now, only pushes to main will trigger a build at all:

```
on:
  push:
    branches:
    - main
```

Compare this with API which has an additional `pull_request` trigger:

```
on:
  push:
    branches:
      - 'main'
    tags:
      - '*'
  pull_request:
```

I don't think the tags are needed for `ui`, but we'll definitely need the `pull_request` trigger that kicks off this workflow, telling us if the image builds correctly. We then skip pushing the image to the image registry using the condition in `push`

tarrow (approved):

PR #880, opened 2024-09-03
Fix logo upload

AndrewKostka:
Bug: T373886


tarrow (approved):
Apparently somewhere in the major axios updates this became necessary

PR #894, opened 2024-09-24
refactor skin selection

deer-wmde:
https://phabricator.wikimedia.org/T375546

~~These changes allow the selection of Vector 2022 and MinervaNeue as default skins for a wiki.~~

- I refactored some logic to separate the displayed strings from the values passed to the API, which allowed me to streamline the skin names how they appear in mediawiki itself
  - this turned out a little bit more complex than I hoped since the v-select component expects an array and not a json object

- Additionally I refactored the simple alerts to the more pleasent looking snackbar alerts as we implemented them in the QuestCaptcha component
  - While I was at it I refactored that part into it's own component instead of copy-pasting the code


src/components/Pages/ManageWiki/Cards/QuestyCaptcha.vue
```diff
@@ -81,24 +81,18 @@
...
-        <v-btn
-          text
-          variant="text"
-          @click="message.show = false"
-        >
-          Close
-        </v-btn>
-      </template>
-    </v-snackbar>
+    <Message ref="message" />
```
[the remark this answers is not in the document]
deer-wmde: nah I'm just a vue noob

src/components/Pages/ManageWiki/Cards/Skin.vue
```diff
@@ -3,15 +3,13 @@
...
       <v-select
-        :items="items"
+        :items="skins"
         label="Skin"
         placeholder="Pick a Skin"
         hint="The default skin is Vector."
         persistent-hint
         prepend-icon="mdi-web"
-        v-model="skin"
-        :disabled="inFlight"
```
deer-wmde, line 13: Are these really unused? Just realized parent components could pass these probably

src/components/Pages/ManageWiki/Cards/QuestyCaptcha.vue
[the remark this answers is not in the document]
deer-wmde: just looked at the docs, I'm not just passing data but calling a method, not sure if there would be another way then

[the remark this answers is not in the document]
deer-wmde: I guess you could argue that we shouldnt know about the method and design the component in a way that changing the properties calls the method, but now I'm way out of bounds of my knowledge about vue architecture

src/components/Pages/ManageWiki/Cards/Skin.vue
```diff
@@ -22,45 +20,62 @@
...
   props: [
     'wikiId'
   ],
   data () {
     return {
-      items: [
-        'Vector',
-        'Modern',
-        'Timeless'
+      skins: [
```
rosalieper, line 42: I would expect `minerva` and `vector-2022`  to be part of this array. As I understood we want them added to the skins

rosalieper (commented):
I tried it locally and did not see more skins

src/components/Pages/ManageWiki/Cards/Skin.vue
deer-wmde: ah right, so we removed them again because they are not ready to be used yet. After our next mediawiki update we will be able to include them here. Until then I changed this PR so that it's just some code refactoring, basically preparation to add them in the future + reuse of the snackbar for alerts

rosalieper: I see I will then merge this

rosalieper (approved):

PR #928, opened 2024-11-18
Update copy for contact page

AndrewKostka:
Bug: [T361588](https://phabricator.wikimedia.org/T361588)

PR #929, opened 2024-11-18
fix: change 'channel' to 'group'

dati18:
Bug: T374791


tarrow (approved):
Nice!

tarrow (commented):
This might be a race with https://github.com/wbstack/ui/pull/928 maybe ping @AndrewKostka so you don't tread on each others toes

PR #934, opened 2024-12-02
Make log in card responsiv for screens as small as 320px

rosalieper:


dati18 (approved):
checked it locally, code-wise and function-wise it LGTM. When Charlie is happy with it then we can ship it

tarrow:
FYI there are some UX changes to be made

dati18 (approved):

PR #940, opened 2025-03-25
update axios package to at least 1.8.2

deer-wmde:
context https://github.com/wbstack/ui/security/dependabot/145


deer-wmde:
closed in favor of #942

PR #956, opened 2025-03-27
WIP: move stuff to rules but this doesn't quite all work right

tarrow:


deer-wmde:
Since shipping https://github.com/wbstack/ui/pull/942 accidentally also shipped https://github.com/wbstack/ui/pull/938, I am merging and deploying this now

PR #982, opened 2025-05-19
Use state for wiki profile card

AndrewKostka:


src/components/Pages/ManageWiki/Cards/Profile.vue
```diff
@@ -137,30 +137,32 @@ export default {
   },
   computed: {
     hasProfile: function () {
-      return !!this.profile.updated_at
+      return !!this.profile?.updated_at
```
tarrow, line 140: ooh optional chaining (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) fancy

tarrow (approved):
lgtm; only played with the "unhappy" path since there is no mock backend for successfully saving the updated profile.

Maybe we want to look at adding that later when we do T394521 ?

tarrow:
Yeah, I went back and checked it worked for me. I also double checked that my Profile was there in the db with by manually poking around with `tinker`. Still all looks good to me and the approval stands; happy to see it merged whenever :)

PR #1012, opened 2025-07-15
Add complaint link and redesign footer

dati18:


rosalieper (approved):

dati18:
I have to remove the logic behind "flexible breakpoints" and fallback to use static CSS breakpoint because it's annoying, waste of effort and totally fragile. 
In Firefox it can work fine and the breakpoint is calculated at480px, but on Chrome it is 450px and weird behaviors occur. This is due to Firefox and Chrome calculate the widths of flex items, scrollbars, subpixel rounding, or default styles differently. In my code, the breakpoint is determined by JavaScript measuring the combined width of the two columns plus a fixed value `(col1Width + col2Width + 176)`. This calculation can yield slightly different results in different browsers

rosalieper:
I wonder why I see already merged changes in your PR. could you try rebasing please? The form is a merged change and it is part of your change.

rosalieper (approved):

PR #1018, opened 2025-08-01
Align font used in the complaint form with the rest of the site

rosalieper:
refactor the css and update the fonts


src/components/Pages/Complaint.vue
```diff
@@ -181,7 +181,7 @@ export default {
   align-self: stretch;
   color: #000;
   font-feature-settings: 'liga' off, 'clig' off;
-  font-family: Lato;
+  font-family: Lato, Segoe UI, sans-serif;
```
outdooracorn, line 184: Should we also make this change on line 174?

outdooracorn (changes requested):
Found some more places to reduce our custom CSS and use Vuetify's defaults! :grin:

Did you see my previous comment about the title and description of this PR?

outdooracorn (approved):
Thank you! :sunglasses:

PR #1029, opened 2025-08-22
Fix contact page heading alignment (Bug: T401820)

Shreya71703:
### Summary
This PR fixes the incorrect heading alignment on the Contact page.

### Changes made
- Corrected the heading style to match design guidelines.
- Ensured proper spacing and consistent font size.

### Bug reference
Bug: T401820 (https://phabricator.wikimedia.org/T401820)

Please review. Thanks!

### DIFF
--- a/src/components/Pages/Complaint.vue
+++ b/src/components/Pages/Complaint.vue
@@ -2,6 +2,13 @@
   <v-main>
     <v-responsive max-width="616px" min-width="250px" >
       <div class="body">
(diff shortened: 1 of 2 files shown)


tarrow:
What is the change to `src/components/Pages/Complaint.vue` about? Is this a result of some attempt to rebase that went wrong? It seems to oddly include some unrelated commit.

I think the only change that should be present is to the contact page.

PR #1042, opened 2025-09-19
Replace Terms of Use checkbox in Instance Creation workflow

rosalieper:


src/components/Cards/CreateWikiWizardStepThree.vue
```diff
@@ -54,32 +54,21 @@
...
-                    Terms of Use</a>
-                </template>
-                Opens in new window
-              </v-tooltip>.
-            </div>
-          </template>
-        </v-checkbox>
+        <div>
+          Previously accepted
+          <v-tooltip bottom>
```
outdooracorn, line 59: Nice tooltip! 👍

src/components/Cards/CreateWikiWizardStepThree.vue
```diff
@@ -54,32 +54,21 @@
...
+                target="_blank"
+                href="/terms-of-use"
+                @click.stop
+                v-on="on"
+              >
+                Terms of Use</a>
+            </template>
+            Opens in new window
+          </v-tooltip> still apply.
+        </div>
```
outdooracorn, line 71: The text here is smaller than the rest of the form and Anton's mockup. A `text-body-1` class on the `div` should do the trick. However, is the text intentionally smaller because it doesn't fit on a single line otherwise? 

Let's ask UX if we should use a smaller font size or can increase the [max-width of this card](https://github.com/wbstack/ui/blob/T404988/src/components/Pages/CreateWiki.vue#L29) to 400px (technically 395px is the minimum it has to be for the text to fit).

outdooracorn (changes requested):
Looking good, I have a couple of comments

src/components/Cards/CreateWikiWizardStepThree.vue
rosalieper: Yes it is. The conclusion was to keep the font smaller. UX prefered the text smaller as long as it fits in one line.

src/components/Cards/CreateWikiWizardStepThree.vue
tarrow: cool, ill click resolve

src/components/Cards/CreateWikiWizardStepThree.vue
tarrow: cool, I'll click resolve :)

PR #1117, opened 2026-04-02
UI: Add hosting policy to wikibase cloud

rosalieper:
Bug: T420471


outdooracorn (approved):
I've pushed a commit that fixes the URLs to what is specified in the ticket.

* Add `target="_blank"` to external links so they open in a new window 
* Make links to `wikibase.cloud` relative so they don't navigate to production from development and staging environments

I've put my grubby hands on this PR now so someone else should probably approve as well!

We've also agreed that we want to make improvements to this page, but that we will do that in a follow-up task: [T422593: Clean up Hosting Policy page on Platform UI](https://phabricator.wikimedia.org/T422593).

tarrow (approved):
additional links changes from @outdooracorn look good to me

PR #1118, opened 2026-04-09
Cleanup custom styling in single-file components

outdooracorn:
* Add `scoped` attribute to `<style>` tags in single-file components
* Use Vuetify-native classes instead of custom styling in `Complaint.vue` and `Interval.vue`
* Remove empty <style> tags in single-file components


dati18:
I found one more empty `<style>` here:
https://github.com/wbstack/ui/blob/main/src/components/Pages/ManageWiki/Cards/Skin.vue

dati18 (approved):

PR #1121, opened 2026-04-13
Improve implementation of Hosting Policy page

rosalieper:
* Remove all custom CSS in favour of native Vuetify2 classes.
* Use `v-alert` instead of custom HTML and CSS.
* Use a color for the `Assessment of size and load` section that show links more clearly.
* Use `v-icon` instead of Unicode emojis so that the icons are consistent across devices.
* Remove incorrect use of `strong`, `em`, and `u` elements in favour of native Vuetify2 classes.

Bug: T422593

### DIFF
--- a/src/components/Pages/HostingPolicy.vue
+++ b/src/components/Pages/HostingPolicy.vue
@@ -8,7 +8,7 @@
           <p>This policy defines Wikimedia Deutschland’s (WMDE) expectations and criteria for hosting on Wikibase Cloud. It explains what kinds of datasets and use cases are supported by WMDE and how those decisions are made and implemented.</p>
           <p>The policy exists to ensure that Wikibase Cloud technical and financial resources, as well as staff capacity are used responsibly, intentionally, in alignment with WMDE’s mission. Wikibase Cloud is not intended to be a generic free hosting service for any purpose. The policy helps us maintain visibility into how the platform is used and provides a framework to support and encourage use cases aligned with the role we want Wikibase Cloud to play in the Wikibase Ecosystem, and to address cases that clearly fall outside that role.</p>
           <h3 class="mb-3 mt-6" id="target-audience">1.1 Target Audience</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>Primary audience: the <a class="font-weight-bold font-italic" href="#review-committee">Review Committee</a> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
             <li>Secondary audience: <span class="font-weight-bold">Wikibase Cloud's current and prospective users</span>, who can use this document to understand the reasoning behind WMDE’s hosting criteria and to self-assess whether their use case is likely to be accepted.</li>
           </ul>
@@ -59,7 +59,7 @@
           <p><span class="font-weight-bold">If after discussion the committee cannot reach a shared decision,</span> this indicates that certain criteria are not defined clearly enough. In such cases, the Review Committee will initiate clarification of the hosting policy before proceeding.</p>
           <h3 class="mb-3" id="prerequisites">3.1 Prerequisites</h3>
           <p>The following conditions must be satisfied before an instance can be hosted on Wikibase Cloud. These conditions are non-negotiable and cannot be offset by positive signals.</p>
-          <h4 class="mb-3" id="required-conditions"><v-icon color="green">mdi-circle</v-icon> Required Conditions (must be true)</h4>
+          <h4 class="mb-3 align-center d-flex" id="required-conditions"><v-icon color="green">mdi-circle</v-icon> Required Conditions (must be true)</h4>
           <p>These conditions must be met for an instance to be eligible for hosting.</p>
           <ul>
             <li>All fields of the <a class="font-italic" href="#wiki-profile">Wiki Profile</a> are complete, providing sufficient detail for the Review Committee to evaluate the intended use.</li>
@@ -69,7 +69,7 @@
             <li>The structured data is in the public domain or released under CC-0.</li>
             <li>The license for the text in other namespaces allows commercial and non-commercial use and derivatives (we recommend using the conformant licenses <a href="https://opendefinition.org/licenses/">from this list</a>).</li>
           </ul>
-          <h4 class="mt-4 mb-3" id="disqualifying-conditions"><v-icon color="red">mdi-circle</v-icon> Disqualifying Conditions (must not be true)</h4>
+          <h4 class="align-center d-flex mt-4 mb-3" id="disqualifying-conditions"><v-icon color="red">mdi-circle</v-icon> Disqualifying Conditions (must not be true)</h4>
           <p>If any of the following conditions apply, the instance cannot be hosted.</p>
           <ul class="mb-4">
             <li>Any indication that the dataset or behaviour may violate or already violates the <a href="/terms-of-use">Terms of Use</a> (e.g., copyright violations, prohibited content, abusive behavior, harassment, privacy violations). <v-icon color="orange">mdi-alert</v-icon> Violations of the Terms of Use are governed exclusively by the Terms of Use and are not subject to the procedures described in this policy.</li>
@@ -99,7 +99,7 @@
               </ul>
             </v-alert>
           </div>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>
               The dataset is only intended to be used by the contributing individual or organization with no clear external audience (people or groups who meaningfully benefit from accessing or reusing this data), including use cases that support purely internal workflows, private organisational processes, or personal documentation, like internal CRM systems, business asset management. <br />The <u>exceptions</u> that do not produce this concerning signal are:
               <ul>
@@ -133,7 +133,7 @@
             <li>take steps to meet the <a class="font-italic" href="#hosting-criteria">Hosting Criteria</a>.</li>
           </ul>
           <h3 class="mb-3" id="review">4.3 Review</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>The <a class="font-italic" href="#review-committee">Review Committee</a>  reviews new applications on an ongoing basis, subject to the availability of its members.</li>
             <li>Each reviewer evaluates the use case independently against the <a class="font-italic" href="#hosting-criteria">Hosting Criteria</a>, noting the situations where essential information is missing or unclear.</li>
             <li>
@@ -192,8 +192,6 @@
                 <li>Is it intended for learning, experimentation, or something else?</li>
               </ul>
             </li>
-          </ul>
-          <ul>
             <li>
               <span class="font-weight-bold">Description of the dataset</span> (for projects intended for sharing and reuse, and for data staging)
               <ul>
@@ -204,8 +202,6 @@
                 <li>Perceived contribution to knowledge equity through data (if any)</li>
               </ul>
             </li>
-          </ul>
-          <ul>
             <li>
               <span class="font-weight-bold">Intended audience</span> (for projects intended for sharing and reuse)
               <ul>
@@ -213,8 +209,6 @@
                 <li>Potential application and tools that will be built on top of it</li>
               </ul>
             </li>
-          </ul>
-          <ul>
             <li>
               <span class="font-weight-bold">Contributor information</span>
               <ul>
@@ -228,12 +222,12 @@
           <h2 class="mt-7 mb-3" id="transparency-and-communication">5. Transparency and Communication</h2>
           <p>We aim to make hosting decisions clear, predictable, and respectful for all applicants.</p>
           <h3 class="mb-3" id="communication-of-decisions-to-applicants">5.1 Communication of Decisions to Applicants</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>Applicants receive a written decision summarizing the Review Committee’s assessment.</li>
             <li>When an instance is not accepted, the decision references relevant criteria, provides the minimum necessary reasoning for suspension without revealing internal deliberations, and any actions required to bring the instance into compliance.</li>
           </ul>
           <h3 class="mb-3" id="public-log-of-decisions">5.2 Public Log of Decisions</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>To support transparency and trust, WMDE will maintain a public log summarizing hosting decisions made by the Review Committee. Its purpose is to help the community understand how the policy is applied in practice.</li>
             <li>Sensitive or private information will not be published.</li>
           </ul>
@@ -245,7 +239,7 @@
           <h2 class="mt-7 mb-3" id="policy-revision-and-updates">6. Policy Revision and Updates</h2>
           <p>To ensure this policy remains accurate, effective, and aligned with WMDE’s mission and operational realities, it will be revised and updated on a regular basis.</p>
           <h3 class="mb-3" id="revision-cycle">6.1 Revision Cycle</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>
               The policy will undergo a formal revision <span class="font-weight-bold">at least once per year</span>, or sooner if:
               <ul>
@@ -257,25 +251,25 @@
             <li>Interim updates may be made if issues are identified during application reviews, re‑evaluations or appeals.</li>
           </ul>
           <h3 class="mb-3" id="responsible-parties">6.2 Responsible Parties</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>The Review Committee performs the policy revision.</li>
             <li>Community members may be consulted on proposed updates, especially when changes affect expectations for contributors or project eligibility.</li>
           </ul>
           <h3 class="mb-3" id="update-process">6.3 Update Process</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>Proposed changes are drafted by the Review Committee based on identified needs, operational learnings, or feedback.</li>
             <li>Draft updates are shared internally and externally for review and alignment.</li>
             <li>Once approved, updates are published on the relevant documentation pages.</li>
             <li>A summary of changes is added to the public decision log, ensuring transparency and traceability.</li>
           </ul>
           <h3 class="mb-3" id="previously-suspended-instances">6.4 Previously Suspended Instances</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>Instances suspended under an earlier version of this policy are not automatically re‑evaluated retrospectively.</li>
             <li>However, users may submit an appeal requesting a new review.</li>
             <li>Appeals for previously suspended instances are assessed according to the updated policy</li>
           </ul>
           <h3 class="mb-3" id="versioning">6.5 Versioning</h3>
-          <ul class="mb-3">
+          <ul class="mb-4">
             <li>Major outdated versions are archived but remain accessible for reference.</li>
           </ul>
         </v-col>
@@ -287,6 +281,5 @@
 <script>
 export default {
   name: 'HostingPolicy',
-  computed: {},
 }
 </script>
--- a/src/components/Pages/HostingPolicy.vue
+++ b/src/components/Pages/HostingPolicy.vue
@@ -101,7 +101,7 @@
           </div>
           <ul class="mb-4">
             <li>
-              The dataset is only intended to be used by the contributing individual or organization with no clear external audience (people or groups who meaningfully benefit from accessing or reusing this data), including use cases that support purely internal workflows, private organisational processes, or personal documentation, like internal CRM systems, business asset management. <br />The <u>exceptions</u> that do not produce this concerning signal are:
+              The dataset is only intended to be used by the contributing individual or organization with no clear external audience (people or groups who meaningfully benefit from accessing or reusing this data), including use cases that support purely internal workflows, private organisational processes, or personal documentation, like internal CRM systems, business asset management. <br />The <span class="text-decoration-underline">exceptions</span> that do not produce this concerning signal are:
               <ul>
                 <li>The instance is a staging ground for a dataset being prepared for later contribution to the other parts of the Wikibase Ecosystem (for example, to Wikidata)</li>
                 <li>The instance supports teaching Wikibase as a software</li>
--- a/src/components/Pages/HostingPolicy.vue
+++ b/src/components/Pages/HostingPolicy.vue
@@ -3,16 +3,16 @@
     <v-container class="fill-height" fluid>
       <v-row justify="center">
         <v-col cols="8" class="policy-content text-body-1">
-          <h1 class="mb-7" id="hosting-policy-for-wikibase-cloud">Hosting Policy for Wikibase Cloud (pilot version)</h1>
-          <h2 class="mb-3" id="purpose-and-scope">1. Purpose and Scope</h2>
+          <h1 id="hosting-policy-for-wikibase-cloud" class="mb-7">Hosting Policy for Wikibase Cloud (pilot version)</h1>
+          <h2  id="purpose-and-scope" class="mb-3">1. Purpose and Scope</h2>
           <p>This policy defines Wikimedia Deutschland’s (WMDE) expectations and criteria for hosting on Wikibase Cloud. It explains what kinds of datasets and use cases are supported by WMDE and how those decisions are made and implemented.</p>
           <p>The policy exists to ensure that Wikibase Cloud technical and financial resources, as well as staff capacity are used responsibly, intentionally, in alignment with WMDE’s mission. Wikibase Cloud is not intended to be a generic free hosting service for any purpose. The policy helps us maintain visibility into how the platform is used and provides a framework to support and encourage use cases aligned with the role we want Wikibase Cloud to play in the Wikibase Ecosystem, and to address cases that clearly fall outside that role.</p>
-          <h3 class="mb-3 mt-6" id="target-audience">1.1 Target Audience</h3>
+          <h3 id="target-audience" class="mb-3 mt-6">1.1 Target Audience</h3>
           <ul class="mb-4">
             <li>Primary audience: the <a class="font-weight-bold font-italic" href="#review-committee">Review Committee</a> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
             <li>Secondary audience: <span class="font-weight-bold">Wikibase Cloud's current and prospective users</span>, who can use this document to understand the reasoning behind WMDE’s hosting criteria and to self-assess whether their use case is likely to be accepted.</li>
           </ul>
-          <h3 class="mb-3" id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
+          <h3 id="relationship-to-other-governance-documents" class="mb-3">1.2 Relationship to other governance documents</h3>
           <p class="font-italic">(These relationships matter to ensure hosting criteria remain coherent with other governance documents, prevent contradictory expectations for contributors, and avoid overlap between responsibilities of WMDE, WMF, and individual communities.)</p>
           <ul class="mb-4">
             <li>This policy complements the <a class="font-weight-bold" href="/terms-of-use">Wikibase Cloud Terms of Use</a>, which define the legal and compliance conditions for using the platform.</li>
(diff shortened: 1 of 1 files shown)


src/components/Pages/HostingPolicy.vue
```diff
@@ -319,37 +312,9 @@ export default {
   margin-bottom: 12px;
 }
```
outdooracorn: We should be able to remove this custom CSS if we do the following:

* `p` elements have a `margin-bottom: 16px` set by default by Vuetify2. I would use that unless there is good reason not to.
* We can use the `mb-3` class on `ul` elements to achieve `margin-bottom: 12px`. However, I would probably use `mb-4` to match `p` elements (unless that looks odd, use your judgement)

src/components/Pages/HostingPolicy.vue
```diff
@@ -271,38 +275,27 @@ export default {
...
-
 .policy-content h1 {
   font-size: 2rem;
-  margin-bottom: 20px;
 }
 
 .policy-content h2 {
   font-size: 1.5rem;
-  margin-top: 28px;
   margin-bottom: 14px;
```
outdooracorn: There is still a lot of custom CSS here modifying margins which can be done via Vuetify-native CSS classes.

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,75 @@
   <v-main>
     <v-container class="fill-height" fluid>
       <v-row justify="center">
-        <v-col cols="8" class="policy-content">
-          <h1 id="hosting-policy-for-wikibase-cloud">Hosting Policy for Wikibase Cloud (pilot version)</h1>
-          <h2 id="purpose-and-scope">1. Purpose and Scope</h2>
+        <v-col cols="8" class="policy-content text-body-1">
+          <h1 class="mb-5" id="hosting-policy-for-wikibase-cloud">Hosting Policy for Wikibase Cloud (pilot version)</h1>
+          <h2 class="mt-7" id="purpose-and-scope">1. Purpose and Scope</h2>
```
outdooracorn: The `mb-5` isn't doing anything here because of the `mt-7` on the `h2` element. See [collapsing margins](https://www.w3.org/TR/CSS2/box.html#collapsing-margins) (and [this SO question](https://stackoverflow.com/questions/19782332/margin-stacks-on-each-other)). I would probably use `mb-X` for both headers.

outdooracorn (changes requested):
Managed to do a quick review.

I would comment out all the custom CSS and see how far you can get with Vuetify-native CSS classes. Reach out for collaboration if that is proving difficult. If we have to add custom CSS we should also write a comment explaining why it was necessary.

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,75 @@
...
+          <h1 class="mb-7" id="hosting-policy-for-wikibase-cloud">Hosting Policy for Wikibase Cloud (pilot version)</h1>
+          <h2 class="mb-3" id="purpose-and-scope">1. Purpose and Scope</h2>
           <p>This policy defines Wikimedia Deutschland’s (WMDE) expectations and criteria for hosting on Wikibase Cloud. It explains what kinds of datasets and use cases are supported by WMDE and how those decisions are made and implemented.</p>
           <p>The policy exists to ensure that Wikibase Cloud technical and financial resources, as well as staff capacity are used responsibly, intentionally, in alignment with WMDE’s mission. Wikibase Cloud is not intended to be a generic free hosting service for any purpose. The policy helps us maintain visibility into how the platform is used and provides a framework to support and encourage use cases aligned with the role we want Wikibase Cloud to play in the Wikibase Ecosystem, and to address cases that clearly fall outside that role.</p>
-          <h3 id="target-audience">1.1 Target Audience</h3>
-          <ul>
+          <h3 class="mb-3 mt-6" id="target-audience">1.1 Target Audience</h3>
+          <ul class="pl-5 mb-3">
```
outdooracorn: Vuetify2 has a `padding-left: 24px` on `ul` elements by default (equivalent to a `pl-6` class). So `pl-5` will reduce the padding by default. Was this intentional?  I'm inclined to just use the defaults.

Let's also use `mb-4` to match the bottom padding of `p` elements.

```suggestion
          <ul class="mb-4">
```

Let's do the same for the other `ul` elements too (except for lists nested in other lists).

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,75 @@
...
             <li>Primary audience: the <strong><em><a href="#review-committee">Review Committee</a></em></strong> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
```
outdooracorn: Let's use CSS classes instead of `strong` and `em`  tags which have other semantic meanings. The text "Review Committee" is not conveying importance or emphasis.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/strong
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/em

```suggestion
            <li>Primary audience: the <a class="font-weight-bold font-italic" href="#review-committee">Review Committee</a> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,75 @@
...
-          <h3 id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
+          <h3 class="mb-3" id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
           <p><em>(These relationships matter to ensure hosting criteria remain coherent with other governance documents, prevent contradictory expectations for contributors, and avoid overlap between responsibilities of WMDE, WMF, and individual communities.)</em></p>
-          <ul>
+          <ul class="pl-5 mb-3">
             <li>This policy complements the <strong><a href="/terms-of-use">Wikibase Cloud Terms of Use</a></strong>, which define the legal and compliance conditions for using the platform.</li>
             <li><strong>Local policies</strong> set by individual communities for their Cloud instances are out of scope for this document; they can operate independently, but within the boundaries defined here.</li>
             <li>In a broader context of the Wikibase ecosystem, it also complements:</li>
-            <ul>
+            <ul class="mt-2 mb-2">
```
outdooracorn: * `mb-2` doesn't have an effect because of the margins on the parent list.
* I don't think we want `mt-2` in this nested list.


```suggestion
            <ul>
```

outdooracorn (changes requested):
Another quick review with some nitpicks. We are getting there :D

src/components/Pages/HostingPolicy.vue
rosalieper: I wanted the design to remain close enough to the current one.  So I tried to stay with the original style where possible. The css I deleted reduced the default `ul` padding

rosalieper: I will do as you suggested

outdooracorn: Thanks, I think we weren't super clear about whether this change had to match the existing CSS exactly. Sorry about that!

You are also right in that a refactor shouldn't change the functionality.

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,19 +2,19 @@
...
-          <h3 id="target-audience">1.1 Target Audience</h3>
-          <ul>
-            <li>Primary audience: the <strong><em><a href="#review-committee">Review Committee</a></em></strong> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
+          <h3 class="mb-3 mt-6" id="target-audience">1.1 Target Audience</h3>
+          <ul class="mb-4">
+            <li>Primary audience: the <a class="font-weight-bold font-italic" href="#review-committee">Review Committee</a> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
             <li>Secondary audience: <strong>Wikibase Cloud's current and prospective users</strong>, who can use this document to understand the reasoning behind WMDE’s hosting criteria and to self-assess whether their use case is likely to be accepted.</li>
```
outdooracorn:
```suggestion
            <li>Secondary audience: <span class="font-weight-bold">Wikibase Cloud's current and prospective users</span>, who can use this document to understand the reasoning behind WMDE’s hosting criteria and to self-assess whether their use case is likely to be accepted.</li>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,19 +2,19 @@
...
           </ul>
-          <h3 id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
+          <h3 class="mb-3" id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
           <p><em>(These relationships matter to ensure hosting criteria remain coherent with other governance documents, prevent contradictory expectations for contributors, and avoid overlap between responsibilities of WMDE, WMF, and individual communities.)</em></p>
```
outdooracorn:
```suggestion
          <p class="font-italic">(These relationships matter to ensure hosting criteria remain coherent with other governance documents, prevent contradictory expectations for contributors, and avoid overlap between responsibilities of WMDE, WMF, and individual communities.)</p>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,19 +2,19 @@
...
-          <ul>
+          <ul class="pl-5 mb-3">
             <li>This policy complements the <strong><a href="/terms-of-use">Wikibase Cloud Terms of Use</a></strong>, which define the legal and compliance conditions for using the platform.</li>
```
outdooracorn:
```suggestion
            <li>This policy complements the <a class="font-weight-bold" href="/terms-of-use">Wikibase Cloud Terms of Use</a>, which define the legal and compliance conditions for using the platform.</li>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,19 +2,19 @@
...
             <li>Secondary audience: <strong>Wikibase Cloud's current and prospective users</strong>, who can use this document to understand the reasoning behind WMDE’s hosting criteria and to self-assess whether their use case is likely to be accepted.</li>
           </ul>
-          <h3 id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
+          <h3 class="mb-3" id="relationship-to-other-governance-documents">1.2 Relationship to other governance documents</h3>
           <p><em>(These relationships matter to ensure hosting criteria remain coherent with other governance documents, prevent contradictory expectations for contributors, and avoid overlap between responsibilities of WMDE, WMF, and individual communities.)</em></p>
-          <ul>
+          <ul class="pl-5 mb-3">
```
outdooracorn: Can we do the same as https://github.com/wbstack/ui/pull/1121#discussion_r3085075001 for all these top level `ul` elements, please?

```suggestion
          <ul class="mb-4">
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,22 +79,26 @@
...
+          <h3 class="mb-3" id="concerning-signals">➖ 3.2 Concerning Signals</h3>
           <p>These signals indicate potential risk or misalignment. They do not automatically lead to rejection, but must be counterbalanced by positive signals.</p>
-          <div class="note warn">
-            <ul class="signal-rule-list">
-              <big><li><strong>2 concerning signals</strong> lead to disqualification.</li>
-              <li><strong>1 concerning signal</strong> must be counterbalanced with 2 positive signals.</li></big>
-            </ul>
+          <div>
+            <v-alert colored-border type="warning" prominent border="left"  elevation="2">
+              <ul class="mb-1">
```
outdooracorn: As this is in an alert and has sufficient margins/padding, I don't think we need to add any more.

```suggestion
              <ul>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,22 +79,26 @@
...
+          <div>
+            <v-alert color="info" text >
+              <p id="assessing-size-and-load"><strong>Assessment of size and load</strong></p>
+              <p>Wikibase Cloud does not operate with fixed numerical limits for dataset size or traffic at this stage. Decisions about size and load will be made empirically, based on observed platform behavior and past incidents. When an instance creates significant strain on the infrastructure, the situation will be investigated, discussed with the instance manager and documented in a public log together with the decision taken.</p>
+              <p>If a high-load use case is meaningful, the Review Committee may assess whether technical improvements and the effort to make them are reasonable. If they are, the use case will only produce a <em><a href="#concerning-signals">Concerning Signal</a></em>. Over time, documented cases will inform future decisions, creating a growing body of precedent. In all cases, instance managers will be contacted and involved in finding a reasonable solution before further action is taken.</p>
```
outdooracorn: Not something you have introduced, but I notice this is two separate paragraphs on https://www.mediawiki.org/wiki/Wikibase/Wikibase.cloud/Hosting_policy_(pilot_version)#%F0%9F%94%B4_Disqualifying_Conditions_(must_not_be_true)

outdooracorn (dismissed):
Sorry, I wasn't very clear that we should remove all the `strong` and `em` elements unless the text is conveying importance or emphasis. Ask yourself, "if I was reading this sentence out loud, would I shout or otherwise emphasizes that part of the sentence?". If not, then the bold and/or italic styling should be done via CSS classes.

I've added some suggestions to provide more examples to go off, but I haven't done them all.

rosalieper:
>"if I was reading this sentence out loud, would I shout or otherwise emphasizes that part of the sentence?"

I went through it and i think i wouldn't shout for any of those text.  probably not emphasize as well so i removed them all and used the css classes from your suggestions.

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,22 +79,27 @@
...
           <p>These signals indicate potential risk or misalignment. They do not automatically lead to rejection, but must be counterbalanced by positive signals.</p>
-          <div class="note warn">
-            <ul class="signal-rule-list">
-              <big><li><strong>2 concerning signals</strong> lead to disqualification.</li>
-              <li><strong>1 concerning signal</strong> must be counterbalanced with 2 positive signals.</li></big>
-            </ul>
+          <div>
+            <v-alert colored-border type="warning" prominent border="left"  elevation="2">
+              <ul>
+                <li><span class="font-weight-bold">2 concerning signals</span>g> lead to disqualification.</li>
```
outdooracorn, line 95: I think some extra characters slipped in here:

```suggestion
                <li><span class="font-weight-bold">2 concerning signals</span> lead to disqualification.</li>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -190,71 +195,71 @@
...
+          <h2 class="mt-7 mb-3" id="policy-revision-and-updates">6. Policy Revision and Updates</h2>
           <p>To ensure this policy remains accurate, effective, and aligned with WMDE’s mission and operational realities, it will be revised and updated on a regular basis.</p>
-          <h3 id="revision-cycle">6.1 Revision Cycle</h3>
-          <ul>
-            <li>The policy will undergo a formal revision <strong>at least once per year</strong>, or sooner if:</li>
-            <ul>
+          <h3 class="mb-3" id="revision-cycle">6.1 Revision Cycle</h3>
+          <ul class="mb-4">
+            <li>The policy will undergo a formal revision <span class="font-weight-bold">at least once per year</span>, or sooner if:</li>
+            <ul class="mb-2">
```
outdooracorn, line 236:
```suggestion
            <ul>
```

outdooracorn (commented):
This is looking good! I've spotted two minor changes to be resolved, but then I'm happy for this to be shown to Anton to make sure he is still happy after these changes, and then merged.

src/components/Pages/HostingPolicy.vue
rosalieper: I think this margin is needed here. I did the same on line 136 and 102. I find it more readable when there is some margin between the list of sub bullet points and the next major bullet point

src/components/Pages/HostingPolicy.vue
```diff
@@ -190,71 +195,71 @@
               <li>Perceived contribution to knowledge equity through data (if any)</li>
             </ul>
           </ul>
```
outdooracorn: Oh, I've just noticed that each of these top level bullet points are a separate list (`ul`). This is semantically different to a single list with multiple list items (`li`).

Also, according to https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ul sub-lists should be a child of an `li` element, not a child of a `ul` element. I think most of the nested lists need updating to address that.

src/components/Pages/HostingPolicy.vue
outdooracorn: Looking at https://www.mediawiki.org/wiki/Wikibase/Wikibase.cloud/Hosting_policy_(pilot_version) there seems to be the same amount of whitespace between a nested list and the parent list item as there is between non-nested list items.

We should do the same here (have the same spacing between all list items regardless of if it's nested or not) unless PM/UX say otherwise.

outdooracorn (changes requested):

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,77 @@
...
-          <h3 id="target-audience">1.1 Target Audience</h3>
-          <ul>
-            <li>Primary audience: the <strong><em><a href="#review-committee">Review Committee</a></em></strong> responsible for reviewing and approving Wikibase Cloud hosting requests for continued use. The policy provides a shared framework for making consistent, transparent decisions grounded in mission alignment, resource realities, and inclusivity goals.</li>
-            <li>Secondary audience: <strong>Wikibase Cloud's current and prospective users</strong>, who can use this document to understand the reasoning behind WMDE’s hosting criteria and to self-assess whether their use case is likely to be accepted.</li>
+          <h3 class="mb-3 mt-6" id="target-audience">1.1 Target Audience</h3>
+          <ul class="mb-3">
```
outdooracorn: How come a load of `ul` elements have changed from `mb-4` to `mb-3`? [We agreed](https://github.com/wbstack/ui/pull/1121#discussion_r3085075001) to use `mb-4` to match the default margins of `p` elements.

Let's be consistent and use `<ul class="mb-4">` for all top level `ul` elements.

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,77 @@
...
-            <li>The licensing statement is visible in the footer. <span class="highlight">Workflows enabling this will be implemented before enforcement of the policy and clear instructions will be documented.</span></li>
+            <li>The licensing statement is visible in the footer. <span class="yellow lighten-4 px-1">Workflows enabling this will be implemented before enforcement of the policy and clear instructions will be documented.</span></li>
             <li>The structured data is in the public domain or released under CC-0.</li>
             <li>The license for the text in other namespaces allows commercial and non-commercial use and derivatives (we recommend using the conformant licenses <a href="https://opendefinition.org/licenses/">from this list</a>).</li>
           </ul>
-          <h4 id="disqualifying-conditions">🔴 Disqualifying Conditions (must not be true)</h4>
+          <h4 class="mt-4 mb-3" id="disqualifying-conditions"><v-icon color="red">mdi-circle</v-icon> Disqualifying Conditions (must not be true)</h4>
```
dati18: Image
The icon and the text are a bit off center. 
Suggestion: check this out  https://github.com/wbstack/ui/blob/f47bebc7173a6f814170c80e3982dd143fd7ebb3/src/components/Pages/TermsOfUse.vue#L275

dati18 (dismissed):

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,201 @@
...
+              <span class="font-weight-bold">Description of the dataset</span> (for projects intended for sharing and reuse, and for data staging)
+              <ul>
+                <li>What kind of data will be hosted? (topic, domain, scope)</li>
+                <li>Rough size and expected growth (if known)</li>
+                <li>Expected contribution workflows (manual, imports, bots)</li>
+                <li>Anticipated approach to taking care of this dataset</li>
+                <li>Perceived contribution to knowledge equity through data (if any)</li>
+              </ul>
+            </li>
           </ul>
```
outdooracorn: These are still separate lists (`ul`) rather than a single list with multiple items.

src/components/Pages/HostingPolicy.vue
outdooracorn: I'm not sure what you are suggesting here @dati18.

Using the `d-flex` and `align-center` classes on the `h4` element is probably the way to go. See https://v2.vuetifyjs.com/en/styles/flex/#flex-align.

outdooracorn (changes requested):

src/components/Pages/HostingPolicy.vue
rosalieper: oh sorry, i think i misunderstood this comment
> Looking at https://www.mediawiki.org/wiki/Wikibase/Wikibase.cloud/Hosting_policy_(pilot_version) there seems to be the same amount of whitespace between a nested list and the parent list item as there is between non-nested list items.


 as though we want the same spacing before and  after a list and sub list like  on the mediawiki page. So this change was an attempt to get that. Now reading that again I understand we want same spacing between list items.

src/components/Pages/HostingPolicy.vue
rosalieper: I think Dat's suggestion is pointing to trying to put the icon and the text as data in a table. while i haven't tried both yet,  I find the `align-center` option cleaner.

dati18: @outdooracorn I suggested a trick to put icon and text in a table. But `align-center` seems to be easier to be done

src/components/Pages/HostingPolicy.vue
```diff
@@ -269,87 +290,3 @@ export default {
   computed: {},
```
dati18: it is empty so you can remove it

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,201 @@
...
-              <li>Brief description of the individual(s), informal group or organization expected to contribute to the project</li>
-              <li>Their mission, economical sector, field of work, approximate size, shared interests that bring them together in this project</li>
-              <li>Any relevant constraints (funding, technical skills, infrastructure)</li>
-              <li>Perceived contribution to knowledge equity through the people holding and sharing the data (if any)</li>
-            </ul>
+            <li>
+              <span class="font-weight-bold">Contributor information</span>
+              <ul>
+                <li>Brief description of the individual(s), informal group or organization expected to contribute to the project</li>
+                <li>Their mission, economical sector, field of work, approximate size, shared interests that bring them together in this project</li>
```
dati18, line 216: I think it should be `economic sector`. "Economical" means cost-effective, right? Like, "A hybrid car is an economical option."

rosalieper: I am not sure. Unless it is a clear typo. I would prefer we (engineers) do not change the actual text content of this page

dati18: you can confirm with Anton :)

outdooracorn: Actually @dati18, please can you take lead on following this up? This PR is about improving the code quality of the hosting policy page. If there are content changes that need to be made they need to be signed off by product and done in a separate PR. Thanks

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,77 @@
...
+          <p><span class="font-weight-bold">If information is insufficient to evaluate an instance,</span> the reviewers request clarification before making a final decision.</p>
+          <p><span class="font-weight-bold">If after discussion the committee cannot reach a shared decision,</span> this indicates that certain criteria are not defined clearly enough. In such cases, the Review Committee will initiate clarification of the hosting policy before proceeding.</p>
+          <h3 class="mb-3" id="prerequisites">3.1 Prerequisites</h3>
           <p>The following conditions must be satisfied before an instance can be hosted on Wikibase Cloud. These conditions are non-negotiable and cannot be offset by positive signals.</p>
-          <h4 id="required-conditions">🟢 Required Conditions (must be true)</h4>
+          <h4 class="mb-3 align-center d-flex" id="required-conditions"><v-icon color="green">mdi-circle</v-icon> Required Conditions (must be true)</h4>
```
outdooracorn: Let's add a small gap between these icons and the heading text. For this and the other icons.

```suggestion
          <h4 class="mb-3 align-center d-flex" id="required-conditions"><v-icon class="mr-1" color="green">mdi-circle</v-icon> Required Conditions (must be true)</h4>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,195 @@
...
-            <ul>
-              <li>The instance is a staging ground for a dataset being prepared for later contribution to the other parts of the Wikibase Ecosystem (for example, to Wikidata)</li>
-              <li>The instance supports teaching Wikibase as a software</li>
-              <li>The instance is used to build and test tools for Wikidata or Wikibase</li>
-            </ul>
+          <ul class="mb-4">
+            <li>
+              The dataset is only intended to be used by the contributing individual or organization with no clear external audience (people or groups who meaningfully benefit from accessing or reusing this data), including use cases that support purely internal workflows, private organisational processes, or personal documentation, like internal CRM systems, business asset management. <br />The <u>exceptions</u> that do not produce this concerning signal are:
```
outdooracorn: Same as not using the `em` and `b` elements, let's replace the `u` element with a `span` and CSS. See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/u for more info.

```suggestion
              The dataset is only intended to be used by the contributing individual or organization with no clear external audience (people or groups who meaningfully benefit from accessing or reusing this data), including use cases that support purely internal workflows, private organisational processes, or personal documentation, like internal CRM systems, business asset management. <br />The <span class="text-decoration-underline">exceptions</span> that do not produce this concerning signal are:
```

outdooracorn (changes requested):
I think we are nearly there! Two small comments.

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,195 @@
...
+              <p>Wikibase Cloud does not operate with fixed numerical limits for dataset size or traffic at this stage. Decisions about size and load will be made empirically, based on observed platform behavior and past incidents. When an instance creates significant strain on the infrastructure, the situation will be investigated, discussed with the instance manager and documented in a public log together with the decision taken.</p>
+              <p>If a high-load use case is meaningful, the Review Committee may assess whether technical improvements and the effort to make them are reasonable. If they are, the use case will only produce a <a class="font-italic" href="#concerning-signals">Concerning Signal</a>.</p>
+              <p>Over time, documented cases will inform future decisions, creating a growing body of precedent. In all cases, instance managers will be contacted and involved in finding a reasonable solution before further action is taken.</p>
+            </v-alert>
           </div>
-          <h3 id="concerning-signals">➖ 3.2 Concerning Signals</h3>
+          <h3 class="mb-3" id="concerning-signals">➖ 3.2 Concerning Signals</h3>
```
outdooracorn: This is still a unicode emoji rather than a `v-icon`. Same for `➕ 3.3 Positive Signals`.

outdooracorn (changes requested):
Latest changes look good 👍. I've noticed two more places than need the Unicode emoji being removed.

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,195 @@
...
+              <p>If a high-load use case is meaningful, the Review Committee may assess whether technical improvements and the effort to make them are reasonable. If they are, the use case will only produce a <a class="font-italic" href="#concerning-signals">Concerning Signal</a>.</p>
+              <p>Over time, documented cases will inform future decisions, creating a growing body of precedent. In all cases, instance managers will be contacted and involved in finding a reasonable solution before further action is taken.</p>
+            </v-alert>
           </div>
-          <h3 id="concerning-signals">➖ 3.2 Concerning Signals</h3>
+          <h3 class="mb-3" id="concerning-signals"><v-icon class="mr-1" color="black">mdi-minus</v-icon>3.2 Concerning Signals</h3>
```
outdooracorn: Let's use the `-thick` icon as it's closer to the Unicode emoji that was used before. If you're going to add the `mr-1` class to the `v-icon` then we should also add the `d-flex` and `align-center` to the `h` element.

```suggestion
          <h3 id="concerning-signals" class="d-flex align-center mb-3"><v-icon class="mr-1" color="black">mdi-minus-thick</v-icon>3.2 Concerning Signals</h3>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,195 @@
...
+                <li>The instance supports teaching Wikibase as a software</li>
+                <li>The instance is used to build and test tools for Wikidata or Wikibase</li>
+              </ul>
+            </li>
             <li>The intended or observed dataset size or read/write load exceeds or approaches excess of current sustainable limits, or the instance exhibits sudden or disproportionate spikes in traffic that strain infrastructure. WMDE decides that allocating additional technical effort to support the use case in the future is justified (see the explanation of how the assessment of size and load happens <a href="#assessing-size-and-load">above</a>).</li>
             <li>Used purely as a MediaWiki, no structured data involved</li>
             <li>Data infested with spammers and other vandalism</li>
           </ul>
-          <h3 id="positive-signals">➕ 3.3 Positive Signals</h3>
+          <h3 class="mb-3" id="positive-signals"><v-icon class="mr-1" color="black">mdi-plus</v-icon>3.3 Positive Signals</h3>
```
outdooracorn: Let's use the `-thick` icon as it's closer to the Unicode emoji that was used before. If you're going to add the `mr-1` class to the `v-icon` then we should also add the `d-flex` and `align-center` to the `h` element.

```suggestion
          <h3 id="positive-signals" class="d-flex align-center mb-3"><v-icon class="mr-1" color="black">mdi-plus-thick</v-icon>3.3 Positive Signals</h3>
```

outdooracorn (approved):
I'm happy with this now. Let's get PM review before merging.

src/components/Pages/HostingPolicy.vue
```diff
@@ -2,75 +2,77 @@
   <v-main>
     <v-container class="fill-height" fluid>
       <v-row justify="center">
-        <v-col cols="8" class="policy-content">
-          <h1 id="hosting-policy-for-wikibase-cloud">Hosting Policy for Wikibase Cloud (pilot version)</h1>
-          <h2 id="purpose-and-scope">1. Purpose and Scope</h2>
+        <v-col cols="8" class="policy-content text-body-1">
+          <h1 id="hosting-policy-for-wikibase-cloud" class="mb-7">Hosting Policy for Wikibase Cloud (pilot version)</h1>
+          <h2  id="purpose-and-scope" class="mb-3">1. Purpose and Scope</h2>
```
outdooracorn:
```suggestion
          <h2 id="purpose-and-scope" class="mb-3">1. Purpose and Scope</h2>
```

src/components/Pages/HostingPolicy.vue
```diff
@@ -79,182 +81,195 @@
...
-            <li>The <em><a href="#review-committee">Review Committee</a></em>  reviews new applications on an ongoing basis, subject to the availability of its members.</li>
-            <li>Each reviewer evaluates the use case independently against the <em><a href="#hosting-criteria">Hosting Criteria</a></em>, noting the situations where essential information is missing or unclear.</li>
-            <li>Reviewers compare their evaluation and decide on the next steps.</li>
-            <ul>
-              <li>If essential information is missing or unclear, reviewers request clarification from the applicant.</li>
-              <li>Applicants may be asked to make adjustments before the review can continue.</li>
-            </ul>
+          <h3 id="review" class="mb-3">4.3 Review</h3>
+          <ul class="mb-4">
+            <li>The <a class="font-italic" href="#review-committee">Review Committee</a>  reviews new applications on an ongoing basis, subject to the availability of its members.</li>
```
outdooracorn:
```suggestion
            <li>The <a class="font-italic" href="#review-committee">Review Committee</a> reviews new applications on an ongoing basis, subject to the availability of its members.</li>
```

outdooracorn (approved):
PM has reviewed the preview and approved

PR #1129, opened 2026-06-01
msw 0.35.0 roll back

deer-wmde:
I tried a bit looking into this MSW issue #1114 #1043
but didn't find a working solution so far

To unblock netlify UI previews now I suggest we roll back the package until we have a working fix. As far as I understand this package only gets used in the local dev setup and the netlify deployments, so I don't think this breaks anything

```
npm install msw@0.35.0 --save-dev
```

update: I was made aware that the netlify "UI pointing to prod" preview links are not affected by this, so we may not want this roll back.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
         "geckodriver": "^3.2.0",
         "lint-staged": "^11.2.3",
         "material-design-icons-iconfont": "^6.1.0",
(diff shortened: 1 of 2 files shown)


deer-wmde:
not needed anymore https://phabricator.wikimedia.org/T399687

PR #1130, opened 2026-06-02
 Add upcoming terms of use page

rosalieper:
Bug: T420575


src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,815 @@
+<template>
+  <v-main>
+    <v-container class="fill-height" fluid>
+      <v-row justify="center">
+        <v-col cols="8">
```
deer-wmde: - I suggest using `cols="12"`, otherwise currently it looks very crammed on small screens
- to use a narrower column layout for bigger screens adding `md="8"` is possible for example

src/components/Pages/TermsOfUseUpcoming.vue
```diff
@@ -0,0 +1,815 @@
...
+            obtain a copy of our technical and organisational measures on request,
+            to do so please send an email to: datenschutz@wikimedia.de.</i>
+          <h4>ANNEX IV</h4>
+          <b>List of sub-processors</b>
+          <b>EXPLANATORY NOTE:</b>
+          <p>This Annex needs to
+            be completed in case of specific authorisation of sub-processors (Clause
+            7.7(a), Option 1). The controller has authorised the use of the
+            following sub-processors:</p>
+          <ol class="pl-8">
```
deer-wmde: I think these ordered list `ol` elements are the ones want to avoid in these legal pages, as they add automatic numbering which is not represented directly here

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/ol

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,815 @@
+<template>
```
deer-wmde: note/idea about the File structure: as we expect we will have a bunch of different versions of these pages my suggestion is to put them in their own directory `Pages/TermsOfUse/`, maybe sensible filenames could be `Current.vue`, `Upcoming.vue`, and once we have more, `version-$N.vue` or something

deer-wmde:
The footnote links I added need the change from here to work https://github.com/wbstack/ui/pull/1131

rosalieper:
The browser test failure is unrelated to this change.

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1158 @@
...
+      <v-row justify="center">
+        <v-col cols="12">
+          <h1>Terms Of Use</h1>
+          <p class="grey--text text--darken-1">last updated on: November 9, 2025</p>
+          <v-alert
+            text
+            color="info"
+          >
+            <div class="text-h6 mb-3 blue--text text--darken-2">
+              What changed from <a class=""><u>previous version</u></a></div>
```
deer-wmde: I think we could already let this point to the current terms of use 
```suggestion
              What changed from <router-link to="/terms-of-use">previous version</router-link></div>
```

second thought/note: I think there is some ongoing discussion about how we want the URLs to the different versions look like. My latest idea was to maybe make use of vue-router aliases: https://v3.router.vuejs.org/guide/essentials/redirect-and-alias.html - but this probably needs further discussion among engineers

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1158 @@
...
+    <v-container class="fill-height" fluid>
+      <v-row justify="center">
+        <v-col cols="12">
+          <h1>Terms Of Use</h1>
+          <p class="grey--text text--darken-1">last updated on: November 9, 2025</p>
+          <v-alert
+            text
+            color="info"
+          >
+            <div class="text-h6 mb-3 blue--text text--darken-2">
```
deer-wmde: It seems currently the color applies only to the heading here, we separate the styling here a bit by wrapping the header in a `p` element or something with the `text-h6 mb-3` classes, while wrapping the whole box into the div with the `blue--text text--darken-2` classes

Also in figma `darken-4` gets used:
Image

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+          <p>
+            <i>Categories of personal data
+              processed</i>
+          <ul class="pl-8">
+            <li>Any categories of personal data contained in the data
+              uploaded by the User.</li>
+          </ul>
+          </p>
+
+          <p>
```
dati18: A paragraph <p> element cannot contain list elements. So when the browser sees a structure like <p> then <ul< inside it, it auto-corrects the DOM by closing the paragraph before the list. That creates a different real DOM than what the template visually suggests.

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+            </li>
+          </ul>
+
+          <h4>Clause 3 &mdash; Third-party beneficiaries</h4>
+          <ul>
+            <li>
+              (a) Data subjects may invoke and enforce these Clauses, as third-party beneficiaries, against the data
+              exporter and/or data importer, with the following exceptions:
+            </li>
+            <ul>
```
dati18: the nested `<ul>` should be a sibling of the parent `<li>` instead of a child of it

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+              (b) In the event that the data importer is in breach of these Clauses or unable to comply with these
+              Clauses, the data exporter shall suspend the transfer of personal data to the data importer until
+              compliance is again ensured or the contract is terminated. This is without prejudice to Clause 14(f).
+            </li>
+            <li>
+              (c) The data exporter shall be entitled to terminate the contract, insofar as it concerns the processing
+              of personal data under these Clauses, where:
+            </li>
+
+            <ul>
```
dati18: the nested `<ul>` should be a sibling of the parent `<li>` instead of a child of it

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+                  </div>
+                </div>
+              </v-expand-transition>
+            </div>
+
+              <div v-if="isMobile" class="mt-2 d-flex justify-space-between">
+                <v-btn
+                  text
+                  class="light-blue--text text-darken-2"
+                  variant="text"
```
dati18: `variant="text"` on the button is Vuetify 3 API style?

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+                    </ul>
+                  </div>
+                </div>
+              </v-expand-transition>
+            </div>
+
+              <div v-if="isMobile" class="mt-2 d-flex justify-space-between">
+                <v-btn
+                  text
+                  class="light-blue--text text-darken-2"
```
dati18: Surrounding components are using `text--darken-*` with 2 dashes and this one is `text-darken-2` with 1 dash. It looks inconsistent for me. Is there a reason for it?

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+                  variant="text"
+                  size="small"
+                  @click="show = !show"
+                >
+                  {{ show ? 'See less' : 'See more' }}
+                </v-btn>
+                <v-btn
```
dati18, line 65: add `label` or `title` that reflects current functionalities (i.e. "see more" or "see less") for this button for screen readers

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+                  class="pa-2 mt-auto d-flex justify-end"
+                  icon
+                  @click="show = !show"
+                >
+                  <v-icon>{{ show ? 'mdi-chevron-up' : 'mdi-chevron-down' }}</v-icon>
+                </v-btn>
+              </div>
+          </v-alert>          <h2>1. Definitions</h2>
```
dati18: The closing `</v-alert>` and the next `<h2>` are on the same line

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+            accordance with Art. 16 DSA Wikimedia provides a mechanism for the User
+            and third parties to report illegal Content hosted on the Platform.</p>
+          <p>10.2 Users are solely responsible for their Content and must ensure
+            compliance with applicable German laws.</p>
+          <p>10.3 The User agrees to indemnify
+            and hold harmless Wikimedia from any claims by third parties arising
+            from Content they publish or actions they take while using our Platform.</p>
+
+          <h2>11. Data Protection</h2>
+          <p>1.11 For comprehensive information on how Wikimedia
```
dati18: it should be section `11.1` not `1.11`

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+            our <router-link to="/privacy-policy">Privacy
+              Policy</router-link>.</p>
+          <p>11.2 To the extent that the Content provided by the
+            User or by contributors and users of the User’s Wikimedia Instance
+            contains personal data of third parties, the User remains the controller
+            of such data and Wikimedia acts as the processor of the User’s personal
+            data under the GDPR. To this end, the Parties enter into the Standard
+            Contractual Clauses (Commission Implementing Decision (EU) 2021/915 of 4
+            June 2021) attached hereto in the Appendix 1.</p>
+          <p>11.3 If and sofar the User
```
dati18: spelling check: "so far" instead of "sofar"

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1162 @@
...
+<style scoped>
+h3 {
+  margin-top: 8px;
+}
+
+h4,
+h5 {
+  margin-top: 4px;
+}
+.card-design {
```
dati18: `.card-design` is defined in `scoped` but not used. You should apply it if required or remove this dead CSS

dati18 (changes requested):
here is my review, please check :)

src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper: image


I checked on the browser and the DOM seem correct and the oder of elements here similar to the vue template.

src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper: you are right. This was a typo

src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper: good catch! thank you.

src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper: I asked the PM on the original doc.

src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper: I am not sure what you want me to change here. the `<ul>` is a child of the parent `<ul>` not a child of `<li>`

src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper: since the btn text is dynamic and visible am not convinced this is necessary.

deer-wmde:
I updated the branch since #1131 got merged and to re-trigger test CI

src/components/Pages/TermsOfUse/Upcoming.vue
deer-wmde: hehe you guys are drilling deep into the HTML specs :D

technically Dat is right as per the specification https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#phrasing_content

practically Rosalie is also right in the sense that the Browser DOM allows it to display without errors anyway, browser engines are actually highly tolerant in terms of how to handle HTML

I think I added this so I will just go ahead and adjust it, since it seems cleaner

src/components/Pages/TermsOfUse/Upcoming.vue
deer-wmde: another instance where Dat is technically right - if I remember correctly I deliberately went against the spec here, because the "right" way to do it here makes the markup imo way more unreadable, and it's hard to see if this is now still correct or if some end tag is missing. So the per-spec way to do it would look like this:

```html
          <ul>
            <li>
              (a) The data importer shall promptly inform the data exporter if it is unable to comply with these
              Clauses, for whatever reason.
            </li>
            <li>
              (b) In the event that the data importer is in breach of these Clauses or unable to comply with these
              Clauses, the data exporter shall suspend the transfer of personal data to the data importer until
              compliance is again ensured or the contract is terminated. This is without prejudice to Clause 14(f).
            </li>
            <li>
              (c) The data exporter shall be entitled to terminate the contract, insofar as it concerns the processing
              of personal data under these Clauses, where:

              <ul>
                <li>
                  (i) the data exporter has suspended the transfer of personal data to the data importer pursuant to
                  paragraph (b) and compliance with these Clauses is not restored within a reasonable time and in any
                  event within one month of suspension;
                </li>
                <li>
                  (ii) the data importer is in substantial or persistent breach of these Clauses; or
                </li>
                <li>
                  (iii) the data importer fails to comply with a binding decision of a competent court or supervisory
                  authority regarding its obligations under these Clauses.
                  In these cases, it shall inform the competent supervisory authority of
                  such non-compliance. Where the contract involves more than two Parties,
                  the data exporter may exercise this right to termination only with
                  respect to the relevant Party, unless the Parties have agreed
                  otherwise.
                </li>
              </ul>
            </li>

            <li>
              (d) Personal data collected by the data exporter in the EU that has been transferred prior to the
              termination of the contract pursuant to paragraph (c) shall immediately be deleted in its entirety,
              including any copy thereof. The data importer shall certify the deletion of the data to the data exporter.
              Until the data is deleted or returned, the data importer shall continue to ensure compliance with these
              Clauses. In case of local laws applicable to the data importer that prohibit the return or deletion of the
              transferred personal data, the data importer warrants that it will continue to ensure compliance with
              these Clauses and will only process the data to the extent and for as long as required under that local
              law.
            </li>
            <li>
              (e) Either Party may revoke its agreement to be bound by these Clauses where (i) the European Commission
              adopts a decision pursuant to Article 45(3) of Regulation (EU) 2016/679 that covers the transfer of
              personal data to which these Clauses apply; or (ii) Regulation (EU) 2016/679 becomes part of the legal
              framework of the country to which the personal data is transferred. This is without prejudice to other
              obligations applying to the processing in question under Regulation (EU) 2016/679.
            </li>
          </ul>
```

Which for me at least personally is very hard to parse when editing this giant markup doc. Super interested in what you guys think about this.

deer-wmde:
looks ready from my side! :ok_hand:

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1160 @@
...
+              insulting, abusive, offensive, racist, threatening, harmful to minors,
+              pornographic, violates personal rights, promotes violence or sedition,
+              incites criminal acts, provides instructions on how to commit criminal
+              acts, or provides services that involve pornographic and/or erotic
+              Content or any other illegal Content</li>
+
+            <li>Upload or distribute Content
+              that is copied as a whole or in part from another protected work or
+              material without the permission of the respective copyright owner;
+              *Upload or distribute Content that violates or infringes the rights of
```
dati18: this should be a new `<li>` block

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1160 @@
...
+  <v-main>
+    <v-container class="fill-height" fluid>
+      <v-row justify="center">
+        <v-col cols="12">
+          <h1>Terms Of Use</h1>
+          <v-alert
+            type="info"
+          >
+            This is an upcoming version. You can find the
+            <router-link class="white--text " to="/terms-of-use">current version here</router-link>
```
dati18: nitpicking: remove the blank space in `class="white--text "`

src/components/Pages/TermsOfUse/Upcoming.vue
```diff
@@ -0,0 +1,1160 @@
...
+
+            <li>Upload or distribute Content that contains personal, confidential or
+              non-public information</li>
+            <li>Contact other users of the Platform in a
+              disrespectful and rude manner, as well as to buy or sell products or
+              services;</li>
+            <li>Provide false data or information, as well as false data or
+              information of third parties; or</li>
+            <li>Sell or otherwise transfer the User Account to another person.</li>
+          </ul></p>
```
dati18: `<p>` elements must not contain block-level elements like `<ul>`. This is invalid HTML and can cause unpredictable layout/rendering. 
Suggestion:
- Split `5.3` into a standalone `<p>` followed by the `<ul>` (without wrapping the `<ul>` in the paragraph).

dati18 (changes requested):
I suggested a few minor changes, you can take a look, resolve conflicts, re-push the branch again to trigger Github Actions checks.

dati18 (approved):
LGTM now

PR #1128, opened 2026-05-28
DSA Info page

deer-wmde:
- **add bilingual DsaInfo Page**
- **hook up DsaInfo page routing**
- **add Footer link**
- **fix scrolling behaviour**

https://phabricator.wikimedia.org/T408254

Bug: T408254

PR #1131, opened 2026-06-02
fix scroll behaviour

deer-wmde:
This change came up with https://github.com/wbstack/ui/pull/1128 and I put it in its own PR because https://github.com/wbstack/ui/pull/1130 also needs this

This fix addresses several things:
1. scroll to the top of the page after following an internal page link
2. making on-page anchor-links (#) work specified with vue router-link

https://phabricator.wikimedia.org/T408254
https://phabricator.wikimedia.org/T420575

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -142,6 +142,18 @@ const router = new Router({
       },
     },
   ],
+  scrollBehavior (to, from, savedPosition) {
+    // brings viewport back to where it was when using Back button
+    if (savedPosition) return savedPosition
+
+    // scroll to in-page router-link anchor if used
+    if (to.hash) {
+      return { selector: to.hash, behavior: 'smooth' }
+    }
+
+    // default: scroll to the top of the page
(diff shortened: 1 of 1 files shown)


rosalieper (approved):
Tried it and it works!

src/router/index.js
```diff
@@ -142,6 +142,18 @@ const router = new Router({
...
+    // brings viewport back to where it was when using Back button
+    if (savedPosition) return savedPosition
+
+    // scroll to in-page router-link anchor if used
+    if (to.hash) {
+      return { selector: to.hash, behavior: 'smooth' }
+    }
+
+    // default: scroll to the top of the page
+    return { x: 0, y: 0, behavior: 'smooth' }
```
dati18, line 155: This is some Vue 3 magic... I cannot find the same thing on v2

deer-wmde: it's part of vue-router https://v3.router.vuejs.org/guide/advanced/scroll-behavior.html

PR #1134, opened 2026-06-08
dev: separate out metal and docker local dev env files

tarrow:
We discovered that the APP_API_URL needs to differ in the baremetal vs docker use case so this adds separate env file for the baremetal usecase and rename the existing one to clarify it's for docker


dati18 (commented):
weird file names, `.env.development.local.baremetal` and `.env.development.loca.docker` may be better?

tarrow:
sure, if you prefer. I have now renamed them

deer-wmde (approved):

dati18 (approved):

PR #1135, opened 2026-06-09
DNM: blindly increase timeouts to 30s

tarrow:


tarrow:
tests still failing. No point in keeping this around

PR #1136, opened 2026-06-09
DNM: reduce log level for wdio

tarrow:


deer-wmde:
ah good thinking, this helps a lot with inspecting the output!

PR #1139, opened 2026-06-10
testing: use production environment for e2e tests

tarrow:
When using the development environment if a request fails the screen is blocked by a helpful debugging message.

I believe the `login` request that happens in the background regularly fails since we updated passport in the platform api. This then blocks clicking any other elements.

The e2e tests rely on the main branch of api to test against which explains why we transition into status tests failing without merging a PR that had failing status tests.

This commit uses the production environment for running the end to end tests. It also adjusts the npm install step. It retains the development environment for installation and uses npm ci[1] rather than install to make this test more deterministic and use the exact versions of packages specified in the lock file.

Finally this commit also uses a plain docker compose --wait and removes the -d since wait already implies detached mode. [2]

[1] https://docs.npmjs.com/cli/v11/commands/npm-ci
[2] https://docs.docker.com/reference/cli/docker/compose/up/

Bug: T428389

### DIFF
--- a/.env.development.test
+++ b/.env.development.test
@@ -1,5 +1,5 @@
 # For running GitHub Actions tests using docker
-NODE_ENV=development
+NODE_ENV=production
 VUE_APP_API_URL=http://host.docker.internal:8082
 VUE_APP_API_MOCK=0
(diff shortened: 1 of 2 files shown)


deer-wmde:
Happy to see this working now :raised_hands:

deer-wmde (approved):

PR #1140, opened 2026-06-16
ToU: navigation panel component

deer-wmde:
I went for a generic component, wrapped in a "Terms of Use" specific one, which can just be included on any of the ToU pages.

Bug: T407632


src/components/Pages/TermsOfUse/Upcoming.vue
rosalieper, line 1: I think you may have accidentally added and commited this whole file

deer-wmde: Ah well for now I just copied a version from #1130 to have some content as a placeholder

Currently in this file, in this PR, the interesting bits are basically only around Line 5-7 where the new component gets used

```
        <v-col cols="11" md="4" order-md="last">
            <TermsOfUseNavigationPanel />
        </v-col>
```

Maybe after we are sure the approach is the right one I can clean it up so it only provides the new components, then it can get integrated in the other PR

rosalieper: right! sorry. You even mentioned that in the PR description

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -0,0 +1,41 @@
...
+  methods: {
+    isCurrentPage (routeName) {
+      return this.$route.name === routeName
+    },
+  },
+}
+
+</script>
+
+<style scoped></style>
```
tarrow: good idea to leave this there (clearly marked as `scoped`) as a service to the future editors of this component

src/components/Pages/TermsOfUse/NavigationLinks.js
tarrow, line 1: I guess this was added as an attempt to remove the list from the NavigationPanel but not currently used?

src/components/Pages/Components/NavigationPanel.vue
```diff
@@ -0,0 +1,41 @@
...
+<script>
+export default {
+  name: 'NavigationPanel',
+  props: {
+    title: String,
+    links: Array,
+  },
+  methods: {
+    isCurrentPage (routeName) {
+      return this.$route.name === routeName
```
tarrow: I think I'd rather we also pass in the current link as a prop to this component rather than have it magically determined it by the route name.

While we just have two hardcoded routes this works but once we are getting the policy list over the wire I don't think we'll be able to rely on the route name. We'll need to look at the params of the route and I think at that point it's probably cleaner for the caller of this component to specify which is the current page.

To try and give an example I think we will need to handle:
- current
- upcoming
- 2029-01-01
- 2022-01-01

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+    <v-expansion-panels v-bind:value="$vuetify.breakpoint.mdAndUp ? 0 : null" v-if="links">
```
tarrow, line 2: Very nice clean vuetify2 way to handle the responsiveness. I love it.

I did have to go hunting to confirm `value` is the [name for prop that controls if the panel is open or closed](https://v2.vuetifyjs.com/en/api/v-expansion-panels/#props) but that's not on you; for me that's not a super clear prop name :D

src/components/Pages/TermsOfUse/NavigationLinks.js
deer-wmde: correct, that was a previous attempt before wrapping the generic component in a more specific one. It worked but felt a bit awkward and then I forgot to delete the file and checked it in, classic^^

src/components/Pages/Components/NavigationPanel.vue
deer-wmde: So when using this component, passing the route path instead of the route name is more favorable you say? https://github.com/wbstack/ui/pull/1140/changes#diff-b107ba54fd78fedebea128629dc9dc517b1ab094d350c0e34d21b60caf919443R15
When thinking about the API that seems cleaner to me, yes

But the Component could still be able to figure out which is the current one itself I think; I'm missing why we would need to tell it manually

tarrow: I'm thinking that there may be a many-one mapping between paths and versions. e.g.
 - `/upcoming`
 - `/2026-06-30`
may both be supposed to render the same in the brower,

This means that the component that uses this navigation panel needs to pass in which item in the list it is showing.

deer-wmde: I think that's something we can cover with redirects/aliases https://v3.router.vuejs.org/guide/essentials/redirect-and-alias.html

edit: but I guess this is about only needing to touch one repo right? Then we'd have to do that routing in the api instead

edit2: on the other hand, we need *some* routing in this repo anyway, seems confusing to put some parts of it in the api

deer-wmde:
Cleaned the PR up to make it mergeable

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -0,0 +1,39 @@
...
+        <v-expansion-panel>
+            <v-expansion-panel-header>{{ title }}</v-expansion-panel-header>
+
+            <v-expansion-panel-content>
+            <v-list class="wrap">
+                <v-list-item v-for="(link) in links" :key="link.routePath">
+                    <v-list-item-content>
+                        <v-list-item-title v-if="link.routePath == currentPath">
```
dati18: Maybe use strict equality `===` to avoid unintended type mismatch?

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -0,0 +1,39 @@
...
+                            {{ link.title }}
+                        </v-list-item-title>
+
+                        <v-list-item-title v-else>
```
dati18, line 14: `<v-list-item-title>` nested inside another `<v-list-item-title>` look kinda weird to me... I think it can break Vuetify's expected list item structure/styling and lead to inconsistent rendering.

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -0,0 +1,39 @@
...
+  props: {
+    title: String,
+    currentPath: String,
+    links: Array,
+  },
+}
+
+</script>
+
+<style scoped></style>
```
dati18: I have different opinion from Tom. I think removing empty `scoped` to reduce noise and add it back it when we really use it?

src/components/Pages/TermsOfUse/TermsOfUseNavigationPanel.vue
```diff
@@ -0,0 +1,27 @@
...
+
+export default {
+  name: 'TermsOfUseNavigationPanel',
+  components: {
+    PolicyNavigationPanel,
+  },
+  data: () => ({
+    termsOfUseLinks: [
+      { title: 'Upcoming Version', routePath: '/terms-of-use/upcoming' },
+      { title: 'April 11th 2022 (current)', routePath: '/terms-of-use' },
```
dati18, line 16: The link title is hard-coded and becomes stale over time. Is it expected?

src/components/Pages/Components/PolicyNavigationPanel.vue
deer-wmde: ahh good spot! :+1:

src/components/Pages/Components/PolicyNavigationPanel.vue
deer-wmde: personally I like it, as it indicates "currently there are no custom styles here, but if wanted, this is how you add them"

but I think you've got a good point! I took a look and realized we seem to not follow this pattern in this repo (meaning, *if* we have the style section our components, then we also use it, otherwise it seems to be omitted)

src/components/Pages/TermsOfUse/TermsOfUseNavigationPanel.vue
deer-wmde: I think currently yes - in the near future this part will come from the API as I understand, but until then we provide the data statically

PR #1147, opened 2026-07-17
add terms-of-use policy renderer component for single ToU routes

deer-wmde:
This PR introduces the TermsOfUseRenderer component which calls the Platform API to render the terms-of-use routes looking like `/terms-of-use/:activeFrom`

- `/terms-of-use/` and `/terms-of-use/upcoming` don't change with this PR/ticket but will also use the TermsOfUseRenderer in upcoming work

https://phabricator.wikimedia.org/T432340


tarrow (approved):
Looks good to me, I made one change to this PR in 859a1f193beec92625a4e699d60c2de97162afb7 

I think it doesn't make sense to add api methods until we need them. Maybe someone could quickly double check they are happy with this removal and then we can go ahead and deploy it.

rosalieper (approved):

PR #1150, opened 2026-07-22
Display policies from API during user registration

AndrewKostka:
Bug: [T430529](https://phabricator.wikimedia.org/T430529)

### DIFF
--- a/src/backend/api.js
+++ b/src/backend/api.js
@@ -53,6 +53,10 @@ export const checkVerified = async () => {
   const { data } = await axios.get('/auth/login')
   return data.verified
 }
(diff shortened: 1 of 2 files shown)


deer-wmde:
I took a quick look, this seems great and to fulfill the criteria of the ticket.

One thing I noticed is that in case there is only a hosting policy existent it will also just show this in the form. But I think we don't need to worry much about this as a) this is very unlikely to happen and b) as I understand it we catch these cases anyway by prompting the user again to agree to any required policies later

deer-wmde (approved):

PR #1151, opened 2026-07-22
Add HostingPolicyRenderer Component

tarrow:
This commit introduces the HostingPolicyRenderer component which calls the Platform API to render the hosting-policy routes looking like /hosting-policy/:activeFrom

It follows a similar pattern to the TermsOfUseRenderer but doesn't include a navigation component since we do not have multiple versions to consider yet.

It also introduces a stub hosting policy content for viewing on staging. This won't be visible on production because there is not a corresponding Policy object in the production DB

This commit also adjusts the TermsOfUseRenderer to determine the vue component to load using the content_vue_file key from the API response rather than by trying to determine it from the active_from date.

Bug: T432340


src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <v-main>
+    <v-container class="fill-height" fluid>
+      <v-row justify="center">
+        <v-col cols="11" md="4" order-md="last">
```
dati18, line 8: should you put `PolicyNavigationPanel` here?

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
```diff
@@ -0,0 +1,62 @@
...
+        const policyType = 'hosting-policy' // TODO read this from component property
+        const activeFrom = this.policyActiveFrom
+
+        const response = await this.$api.policyByDate({ policyType, activeFrom })
+
+        const metadata = await response.metadata
+        this.policy = versions[metadata.content_vue_file]
+      } catch (error) {
+        console.error(error)
+        alert('Failed to load policy.')
```
dati18: The error message is very generic, and `alert()` is a poor UX for a Vue/Vuetify app. Should we use `snackbar` like other components?

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
```diff
@@ -0,0 +1,62 @@
...
+  methods: {
+    async loadPolicy () {
+      try {
+        const policyType = 'hosting-policy' // TODO read this from component property
+        const activeFrom = this.policyActiveFrom
+
+        const response = await this.$api.policyByDate({ policyType, activeFrom })
+
+        const metadata = await response.metadata
+        this.policy = versions[metadata.content_vue_file]
```
dati18: I'm wondering if `content_vue_file` is empty, then `this.policy` becomes `undefined` and the UI will silently render nothing. Should we `throw` or show an error when key is unknown?

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
dati18: I tested and cannot load `version1` and was hit with this alert instead. Did I do something wrong?

Image

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
tarrow: > It follows a similar pattern to the TermsOfUseRenderer but doesn't include a navigation component since we do not have multiple versions to consider yet.

I tried to explain in the description withheld it for now because there doesn't seem to be multiple versions to consider. We could consider moving the `pilot` into this but I'm not sure it's worth bothering.

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
tarrow: > This won't be visible on production because there is not a corresponding Policy object in the production DB

I think it won't be possible to see something on production because there isn't a hosting policy in the DB there

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
tarrow: yeah, not a bad idea. I'll look at incorporating this with the above comment

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
tarrow: sure, I think we could do something better. Perhaps we ought to show something inline instead.

deer-wmde:
apart from the already noted points this looks good to me

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
deer-wmde: I addressed this now in b1726a66cbd3bf982da43721952ebdbd2c5ffd13

tarrow: great, I'll resolve this

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
tarrow: resolved by @deer-wmde

tarrow:
Also looks fine to me, I guess we should ship this.

PR #1154, opened 2026-07-24
integrate policy API

deer-wmde:
- Policy pages are now API-driven
- The current/upcoming Terms of Use behavior is centralized in the renderer
- Hosting Policy has been brought into the same rendering path
- Current hosting policy is redirected to pilot version, will be removed later

Bug: T432965


deer-wmde:
@tarrow @dati18 instead of duplicating the logic from the TermsOfUseNavigationPanel for the hosting policy, I just moved it to its own general component because its code was already quite generalized, this seemed like the quicker win from where the code already was

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -25,12 +25,143 @@
...
+        const isCurrentPolicy = currentPolicy && policy.metadata.policy_id === currentPolicy.metadata.policy_id
+        const isUpcomingPolicy = upcomingPolicy && policy.metadata.policy_id === upcomingPolicy.metadata.policy_id
+
+        return {
+          routePath: this.routePathForPolicy({ activeFrom, isCurrentPolicy, isUpcomingPolicy }),
+          title: this.titleForPolicy({ activeFrom, isCurrentPolicy, isUpcomingPolicy }),
+        }
+      })
+    },
+    currentLink: function () {
```
tarrow: I believe we're overloading the term `current` here.

Both to mean "the latest, active version of a given policy type on the platform" and "the one the renderer is showing right now at this route".

I wish I had a good suggestion for name here (which I believe covers the second case from above). Maybe `presentPosition` or `entryWithoutLink`?

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
```diff
@@ -4,8 +4,18 @@
       An error occurred while trying to load the requested policy. Please try again later.
     </v-alert>
     <v-container class="fill-height" fluid v-if="!error">
+      <v-row v-if="isUpcomingPolicy" justify="center">
```
tarrow: In the ToU renderer we removed this `row` etc. Was there a reason to do it differently or just oversight? I'm actually indifferent as to which we pick but I think consistency wouldn't be a good idea

src/components/Pages/Components/PolicyNavigationPanel.vue
```diff
@@ -25,12 +25,143 @@
...
 <script>
+const parseActiveFrom = (activeFrom) => {
+  if (activeFrom === null || activeFrom === undefined) {
+    return null
+  }
+
+  return new Date(`${activeFrom}T00:00:00Z`)
+}
+
+const comparePolicies = (left, right) => {
```
tarrow: this `sort` and comparison logic is all very pretty.

I think it would be way clearer if we had a working test suite where we could see a few of these all nicely rendered but I'm not sure now if the time to do that unless it seems very easy

src/components/Pages/Components/PolicyNavigationPanel.vue
tarrow: We looked at this and used words like "rendered" and "currently rendered"

src/components/Pages/Components/PolicyNavigationPanel.vue
tarrow: Didn't put together a test suite

src/components/Pages/HostingPolicy/HostingPolicyRenderer.vue
tarrow: we decided to use the "no row" syntax

dati18 (approved):

PR #1155, opened 2026-07-27
Fix duplicate key error on registration page

AndrewKostka:
Bug: [T433220](https://phabricator.wikimedia.org/T433220)


rosalieper (approved):
I tried it and it works

PR #1160, opened 2026-07-29
Add flow for existing users to accept new policies

AndrewKostka:
Bug: [T401223](https://phabricator.wikimedia.org/T401223)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,10 +1,14 @@
 <template>
   <v-app id="app">
+      <ErrorSnackBar/>
       <Navbar></Navbar>
-      <router-view v-if="customLayout"/>
-      <v-container v-else class="full-height-content">
-        <router-view></router-view>
-      </v-container>
+      <div v-if="!failedToGetMissingPolicies" class="content">
+        <PolicyAcceptanceModal v-if="hasMissingPolicies"/>
+        <router-view v-if="customLayout"/>
+        <v-container v-else class="full-height-content">
+          <router-view></router-view>
+        </v-container>
+      </div>
       <Foot :class="['padding-footer', { 'tall-footer': customLayout }]"></Foot>
       <Interval
         v-if="this.$store.getters.isLoggedIn && !this.$store.getters.currentUser.verified"
@@ -18,6 +22,8 @@
 import Navbar from '@/components/Layout/Navbar'
 import Foot from '@/components/Layout/Foot'
 import Interval from '@/components/Util/Interval'
+import PolicyAcceptanceModal from '@/components/Modals/PolicyAcceptanceModal'
+import ErrorSnackBar from '@/components/Components/ErrorSnackBar'
 
 export default {
   name: 'App',
@@ -25,10 +31,20 @@ export default {
     Navbar,
     Foot,
     Interval,
+    PolicyAcceptanceModal,
+    ErrorSnackBar,
   },
   computed: {
     customLayout: function () {
       return this.$route.meta.customLayout
+    },
+    hasMissingPolicies: function () {
+      return this.$route.meta.requiresAuth &&
+        !this.$route.meta.requiresAuth.excludeFromPolicyChecks &&
+        this.$store.getters.missingPolicies?.length
+    },
+    failedToGetMissingPolicies: function () {
+      return this.$store.getters.failedToGetMissingPolicies
     },
   },
   methods: {
@@ -48,10 +64,15 @@ export default {
 .full-height-content {
   height: 100%
 }
-.padding-footer.footer {
-  margin-top: 56px;
-}
 .tall-footer.footer {
   height: 100%;
+}
+</style>
+
+<style scoped>
+.content {
+  position: relative;
+  height: 100%;
+  padding-bottom: 56px;
 }
 </style>
--- a/src/backend/api.js
+++ b/src/backend/api.js
@@ -57,6 +57,15 @@ export const getCurrentPolicies = async () => {
   const { data } = await axios.get('/v1/policies/current')
   return data.items
 }
+export const getMissingPolicies = async () => {
+  const { data } = await axios.get('/v1/policies/missing')
+  return data.items
+}
+export const acceptPolicies = async (policyIds) => {
+  await axios.put('/v1/policy_acceptances', {
+    policy_ids: policyIds,
+  })
+}
 
 /* Wiki endpoints */
 export const countWikis = async () => (await axios.get('/wiki/count')).data.data // TODO This doesn't seem to exist and not used?
--- a/src/components/Modals/PolicyAcceptanceModal.vue
+++ b/src/components/Modals/PolicyAcceptanceModal.vue
@@ -1,0 +1,100 @@
+<template>
+  <div class="overlay">
+    <v-card max-width="580">
+      <v-card-title>
+        Updated policies
+      </v-card-title>
+      <v-card-text>
+        <p class="text-body-1">
+          Please accept the updated
+          <PolicyList
+            :policies="policies"
+            article=""
+            :clickable="false"
+          />
+          to continue using Wikibase Cloud.
+          If you prefer not to keep your account, you can contact us to delete it.
+        </p>
+        <v-checkbox
+          v-model="accepted"
+          hide-details
+          class="mt-2"
+          :disabled="inFlight"
+        >
+          <template v-slot:label>
+            <div class="text-body-1">
+              I agree to the
+              <PolicyList :policies="policies"/>
+              <span>.</span>
+            </div>
+          </template>
+        </v-checkbox>
+      </v-card-text>
+      <v-card-actions>
+        <v-btn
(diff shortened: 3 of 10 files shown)


src/backend/api.js
```diff
@@ -57,6 +57,15 @@ export const getCurrentPolicies = async () => {
   const { data } = await axios.get('/v1/policies/current')
   return data.items
 }
+export const getMissingPolicies = async () => {
```
dati18, line 60: `getMissingPolicies` and `getCurrentPolicies` looks similar (same response shape), only the path differs.
You can consider writing `getPolicies(path)` or something like `getPoliciesByStatus('missing' |'current')` ti reduce repetition.

src/backend/api.js
```diff
@@ -57,6 +57,15 @@ export const getCurrentPolicies = async () => {
...
+  const { data } = await axios.get('/v1/policies/missing')
+  return data.items
+}
+export const acceptPolicies = async (policyIds) => {
```
dati18, line 64: `acceptPolicies` doesn't return anything. It can return `response.data` or something

src/store/error.js
```diff
@@ -0,0 +1,41 @@
+const getDefaultState = () => {
+  return {
+    latestError: null,
+    latestErrorMessage: null,
+    latestErrorDismissible: true,
+  }
+}
+
+const state = getDefaultState()
```
dati18, line 9: exporting a singleton `state` object is a bit off to me. Does it lead to a shared state across store instances?

src/store/error.js
```diff
@@ -0,0 +1,41 @@
...
+
+const getters = {
+  hasLatestError: state => !!state.latestError,
+  latestError: state => state.latestError,
```
dati18, line 13: The action name `latestError` conflicts conceptually with the getter `latestError`, which makes usage and debugging confusing (`dispatch('latestError')` vs `getters.latestError`). Rename the action to something imperative like `reportLatestError` or `setLatestError` to clarify intent.

src/store/policies.js
```diff
@@ -0,0 +1,57 @@
...
+    state.failedToGetMissingPolicies = failed
+  },
+}
+
+const actions = {
+  async missingPolicies ({ dispatch, commit }) {
+    try {
+      commit('setMissingPolicies', await api.getMissingPolicies())
+    } catch (error) {
+      commit('failedToGetMissingPolicies', true)
```
dati18, line 32: The `failedToGetMissingPolicies` flag is set to `true` on error. But if I understand correctly, it is never reset to `false` on succesful fetch?

src/store/policies.js
```diff
@@ -0,0 +1,57 @@
+import { api } from './../backend'
+
+const getDefaultState = () => {
+  return {
+    missingPolicies: null,
+    failedToGetMissingPolicies: false,
+  }
+}
+
+const state = getDefaultState()
```
dati18, line 10: same problem above: exporting singleton state object

src/store/policies.js
```diff
@@ -0,0 +1,57 @@
...
+
+const getters = {
+  missingPolicies: state => state.missingPolicies,
```
dati18, line 13: Consider renaming the action to something verb-based like `fetchMissingPolicies` for clarification.
Having both a getter and an action named `missingPolicies` is confusing (`dispatch('missingPolicies')` vs `getters.missingPolicies`)

src/store/policies.js
```diff
@@ -0,0 +1,57 @@
...
+  },
+}
+
+const actions = {
+  async missingPolicies ({ dispatch, commit }) {
+    try {
+      commit('setMissingPolicies', await api.getMissingPolicies())
+    } catch (error) {
+      commit('failedToGetMissingPolicies', true)
+      dispatch('latestError', {
```
dati18: The `dispatch('latestError', ...)` call is not awaited or returned. Consider `return await dispatch('latestError', ...)` (or at least `await`)

src/App.vue
```diff
@@ -18,18 +22,30 @@
...
+    ErrorSnackBar,
   },
   computed: {
     customLayout: function () {
       return this.$route.meta.customLayout
     },
+    hasMissingPolicies: function () {
+      return this.$route.meta.requiresAuth &&
+        !this.$route.meta.requiresAuth.excludeFromPolicyChecks &&
+        this.$store.getters.missingPolicies?.length
```
dati18, line 44: I don't really get this bit...
`hasMissingPolicies` returns `missingPolicies?.length`, which is a number (or worst case `undefined`). Since this is used in `v-if`, is it less error-prone to return an explicit boolean (for example, `(this.$store.getters.missingPolicies?.length ?? 0) > 0`)?

src/components/Modals/PolicyAcceptanceModal.vue
```diff
@@ -0,0 +1,100 @@
...
+  computed: {
+    policies () {
+      return this.$store.getters.missingPolicies
+    },
+  },
+  methods: {
+    onContactForm () {
+      this.$router.push('/contact')
+    },
+    async onAccept () {
```
dati18, line 76: `inFlight` is set back to `false` only on the success path. If `acceptPolicies` throws/rejects or fails in general, `inFlight` will remain `true` and the UI will be stuck.
`try/catch/finally` can wrap the `dispatch` so `inFlight` is always reset

src/components/Modals/PolicyAcceptanceModal.vue
```diff
@@ -0,0 +1,100 @@
...
+    <v-card max-width="580">
+      <v-card-title>
+        Updated policies
+      </v-card-title>
+      <v-card-text>
+        <p class="text-body-1">
+          Please accept the updated
+          <PolicyList
+            :policies="policies"
+            article=""
```
dati18, line 12: `article` is empty, if the prop is optional, omit it

src/components/Modals/PolicyAcceptanceModal.vue
```diff
@@ -0,0 +1,100 @@
...
+    },
+    async onAccept () {
+      this.inFlight = true
+      const policyIds = Array.from(
+        this.policies,
+        policy => policy.metadata.policy_id,
+      )
+      await this.$store.dispatch('acceptPolicies', policyIds)
+      this.inFlight = false
+    },
```
dati18, line 84: _Optional:_
Add tests for the modal actions: 
1. clicking “Contact Form” navigates to `/contact`
2. “Continue” is disabled until the checkbox is checked
3. `onAccept` re-enables the UI even when `acceptPolicies` rejects (error path).

src/backend/api.js
AndrewKostka: Each represents a different API route in back-end, so I'd prefer to keep them apart here as well.

src/backend/api.js
AndrewKostka: That API route only returns a generic `{'success' => true}` in the normal flow, so there isn't really anything useful to extract from the response data.

src/components/Modals/PolicyAcceptanceModal.vue
AndrewKostka: Here we're specifically overriding its default value of "the " with an empty string.

src/components/Modals/PolicyAcceptanceModal.vue
AndrewKostka: There's a `try/catch` upstream in the store that's supposed to handle any unhappy paths.

src/components/Modals/PolicyAcceptanceModal.vue
AndrewKostka: I think we're no longer writing browser tests for this UI.

src/store/error.js
AndrewKostka: AFAIK, this `state` object is required in "./index.js" for creating the Vuex store.

src/store/error.js
AndrewKostka: Sounds good, will do.

src/store/policies.js
AndrewKostka: AFAIK, this `state` object is required in "./index.js" for creating the Vuex store.

src/store/policies.js
AndrewKostka: Sounds good, will do.

src/store/policies.js
AndrewKostka: When this error occurs, we force the user to reload the page. That'll automatically reset this flag. I'll add a comment for this.

src/store/policies.js
AndrewKostka: That's okay, these can be called asynchronously.

src/App.vue
AndrewKostka: `v-if` relies on a truthy value. That'll only happen here if the array has at least one element. `undefined` and `0` are both falsy.

dati18 (approved):

